PR #15, opened 2019-04-25
Added Write Lifetime Hint support in classifier

micrakow:
Adding write lifetime hint support in classifier, available from kernel 4.13.


modules/cas_cache/classifier.c
```diff
@@ -32,6 +32,10 @@
 #define CAS_CLS_DEBUG_TRACE(format, ...)
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 13, 0)
```
michalwy: Please move this to linux_kernel_version.h

PR #25, opened 2019-05-24
Extending 'configure' script

mmichal10:
Functions and macros dependent on different kernel versions are now generated
before compilation basing on current kernel capabilities instead of hardcoding
them for specific kernels.


configure
```diff
@@ -67,40 +67,514 @@ kernel_not_supp_fail() {
...
+elif compile_module "REQ_WRITE" "linux/blk_types.h"
+then
+	add_define "RQ_DATA_DIR_WR\\" "\tREQ_WRITE"
+fi
+
+if compile_module "REQ_PREFLUSH" "linux/blk_types.h"
+then
+	add_define "CAS_REQ_FLUSH\\" "\tREQ_PREFLUSH"
+	add_define "CAS_FLUSH_SUPPORTED\\" "\t1"
 else
```
micrakow: Please change to elif

configure.d/conf_framework
```diff
@@ -0,0 +1,67 @@
...
+
+compile_module(){
+	if [ $# -gt 1 ]
+	then
+		INCLUDE="#include <$2>"
+	else
+		INCLUDE=""
+	fi
+
+############# TEST MODULE #############
```
arutk: you should be able to indent this entire block of code with tabs if you replace "<<" with "<<-".

configure.d/conf_framework
```diff
@@ -0,0 +1,67 @@
...
+add_define() {
+   echo -n "#define " >> $DEFINE_FILE
+   for arg in "$@"; do
+       echo -e "$arg" >> $DEFINE_FILE
+   done
+   echo "" >> $DEFINE_FILE
+}
+
+add_function() {
+	for arg in "$@"; do
```
arutk: This function can be improved significantly. 

1. Replace 'for' loop with 
printf $1 >> $DEFINE_FILE
2. At the end of framework file override internal field separator
IFS='*'
3. Now you can specify functions in plain text, like this:
add_function "
           int fun(int art)
           {
             ...
           }"

PR #58, opened 2019-07-02
Initialize structures with 0.

mmichal10:
If 'open_cores' flag is not explicite set, cores migth not be added on cache
load.

### DIFF
--- a/modules/cas_cache/service_ui_ioctl.c
+++ b/modules/cas_cache/service_ui_ioctl.c
@@ -61,7 +61,7 @@ long cas_service_ioctl_ctrl(struct file *filp, unsigned int cmd,
 		struct kcas_start_cache *cmd_info;
 		struct ocf_mngt_cache_config cfg;
 		struct ocf_mngt_cache_device_config device_cfg;
-		struct atomic_dev_params atomic_params = { 0 };
+		struct atomic_dev_params atomic_params;
 
 		GET_CMD_INFO(cmd_info, arg);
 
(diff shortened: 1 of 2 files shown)


modules/cas_cache/service_ui_ioctl.c
```diff
@@ -61,9 +61,12 @@ long cas_service_ioctl_ctrl(struct file *filp, unsigned int cmd,
 		struct kcas_start_cache *cmd_info;
 		struct ocf_mngt_cache_config cfg;
 		struct ocf_mngt_cache_device_config device_cfg;
-		struct atomic_dev_params atomic_params = { 0 };
+		struct atomic_dev_params atomic_params;
 
 		GET_CMD_INFO(cmd_info, arg);
+		memset(&cfg, 0, sizeof(cfg));
```
arutk: Looks like this is now memset to 0 twice - the second one is in cache_mngt_prepare_cache_cfg. I suppose the original idea was to zero the structure inside "prepare_cfg" - I suggest to stick to this for all three structures.

modules/cas_cache/service_ui_ioctl.c
```diff
@@ -102,6 +105,7 @@ long cas_service_ioctl_ctrl(struct file *filp, unsigned int cmd,
 		struct ocf_mngt_core_config cfg;
 
 		GET_CMD_INFO(cmd_info, arg);
+		memset(&cfg, 0, sizeof(cfg));
```
arutk: Same comment here.

PR #63, opened 2019-07-08
Check for existing metadata

mmichal10:


casadm/extended_err_msg.c
```diff
@@ -132,6 +132,12 @@ struct {
 		OCF_ERR_NO_METADATA,
 		"No metadata found on device"
 	},
+	{
+		OCF_ERR_METADATA_FOUND,
+		"Old metadata found on device.\nPlease start cache using "
```
arutk: Let's be more verbose here. I suggest something along these lines:

Please load cache metadata using --load option or use --force to discard on-disk metadata and start fresh cache instance.

PR #68, opened 2019-07-16
Fix error handling in case of fatal cache stop error

micrakow:
In case of error during cache stop exported object is not being restored properly.
This may result in bug in case of calling stop once more.

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -1565,13 +1565,11 @@ out:
 }
 
 /**
- * @brief routine implements --remove-cache command.
- * @param[in] device caching device to be removed
+ * @brief routine implements --stop-cache command.
+ * @param[in] cache_id caching device id to be removed
  * @param[in] flush Boolean: shall we flush dirty data before removing cache.
  *		if yes, flushing may still be interrupted by user (in which case
  *		device won't be actually removed and error will be returned)
- * @param[in] allow_interruption shall we allow interruption of dirty
- *		data flushing
  */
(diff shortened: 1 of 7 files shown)


micrakow:
retest this please

KamilLepek:
retest this please

micrakow:
retest this please

robertbaldyga:
Why don't we make `block_dev_destroy_all_exported_objects()` safe to call even if objects have been previously destroyed? This seems to be simpler and more reliable solution, as recreating exported objects may fail, and then we are again exposed to the same issue we are trying to fix.

modules/cas_cache/volume/vol_block_dev_top.c
```diff
@@ -915,6 +925,15 @@ int block_dev_create_exported_object(ocf_core_t core)
 	return result;
 }
 
+static int _block_dev_create_exported_object_visitor(ocf_core_t core, void *cntx){
```
robertbaldyga: Coding style: move function opening brackets to new line.

modules/cas_cache/layer_cache_management.c
```diff
@@ -1658,6 +1658,24 @@ int cache_mngt_exit_instance(ocf_cache_id_t id, int flush)
 	ocf_queue_put(cache_priv->mngt_queue);
 	vfree(cache_priv);
 
+	ocf_mngt_cache_unlock(cache);
+	ocf_mngt_cache_put(cache);
+
+	return status;
+
+create_exp_objs:
```
robertbaldyga: Let's rename it to `restore_exp_obj`.

PR #82, opened 2019-08-20
Reset particular core stats only if core was specified by user.

mmichal10:


mmichal10:
retest this please

PR #97, opened 2019-09-10
Stats refactor

mmichal10:


casadm/statistics_model.c
```diff
@@ -220,238 +217,180 @@ static void print_usage_header(FILE* outfile)
...
+
+	print_val_perc_table_section(outfile, "Total requests",
+				     UNIT_REQUESTS, stats->total.fraction, "%lu",
+					 stats->total.value);
+}
+
+#define get_stat_name(__dst, __len, __name, __postfix) \
+	memset(stat_name, 0, max_stat_len); \
+	strcpy(stat_name, __name); \
+	strcat(stat_name, postfix);
```
imjfckm: Please use safe functions like `strlcpy` and `strlcat`, especially since you have length handy

modules/cas_cache/layer_cache_management.c
```diff
@@ -1744,6 +1744,56 @@ int cache_mngt_interrupt_flushing(const char *cache_name)
...
+		if (result)
+			goto unlock;
+
+		result = ocf_stats_collect_core(core, &stats->usage, &stats->req,
+				&stats->blocks, &stats->errors);
+		if (result)
+			goto unlock;
+
+	} else {
+		result = get_core_by_id(cache, stats->core_id, &core);
```
arutk: So core_id is potentially INVALID here, and you handle this by inspecting return value from get_core_by_id? I think it would be better to chceck core_id vs INVALID and then either get the core and get core stats OR call ocf_stats_collect_part_cache.

modules/cas_cache/layer_cache_management.c
```diff
@@ -1881,6 +1918,36 @@ int cache_mngt_get_core_info(struct kcas_core_info *info)
 	return result;
 }
 
+int cache_mngt_get_core_stats(struct kcas_get_stats *stats)
```
arutk: looks like this is not used

modules/cas_cache/layer_cache_management.c
```diff
@@ -1744,6 +1744,56 @@ int cache_mngt_interrupt_flushing(const char *cache_name)
...
+		result = ocf_stats_collect_core(core, &stats->usage, &stats->req,
+				&stats->blocks, &stats->errors);
+		if (result)
+			goto unlock;
+
+	} else {
+		result = get_core_by_id(cache, stats->core_id, &core);
+
+		if (result) {
+			result = ocf_stats_collect_part_cache(cache, stats->part_id,
```
imjfckm: Shouldn't we just give cache stats only if core_id is equal to `OCF_CORE_ID_INVALID`, not when it's just invalid (i.e. not added)?

modules/cas_cache/layer_cache_management.c
```diff
@@ -1753,6 +1753,58 @@ int cache_mngt_interrupt_flushing(const char *cache_name)
...
+		if (result)
+			goto unlock;
+
+		result = ocf_stats_collect_core(core, &stats->usage, &stats->req,
+				&stats->blocks, &stats->errors);
+		if (result)
+			goto unlock;
+
+	} else {
+		if (stats->core_id == OCF_CORE_ID_INVALID) {
```
imjfckm: This could be in the upper level conditional just as another `else if` not nested here in last else clause

modules/cas_cache/layer_cache_management.c
```diff
@@ -1753,6 +1753,58 @@ int cache_mngt_interrupt_flushing(const char *cache_name)
...
+			result = ocf_stats_collect_part_cache(cache, stats->part_id,
+					&stats->usage, &stats->req, &stats->blocks);
+		} else {
```
imjfckm: Ditto, this coul be just the last `else` clause of the upper level conditional

modules/cas_cache/layer_cache_management.c
mmichal10: It could have upper indentation level, but I left it like this on purpose. Indentation level matches with code logic.

mmichal10:
retest this please

KamilLepek:

PR #99, opened 2019-09-10
cas_disk: Add mandatory blk_mq_ops filed required by some kernels

micrakow:

### DIFF
--- a/configure.d/1_blk_mq.conf
+++ b/configure.d/1_blk_mq.conf
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright(c) 2012-2019 Intel Corporation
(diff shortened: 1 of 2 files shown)


configure.d/1_blk_mq.conf
```diff
@@ -0,0 +1,32 @@
...
+		echo $cur_name "2" >> $config_file_path
+    fi
+}
+
+apply() {
+    case "$1" in
+    "1")
+		add_define "CAS_BLK_MQ_OPS_MAP_QUEUE" ;;
+    "2")
+    		return 0 ;;
```
robertbaldyga: just `return 0;` in header file? does it even compile?

arutk: This is just return code from bash function, it does not get printed to output file.

PR #120, opened 2019-09-25
Fix env_strncpy.

mmichal10:
strlcpy() requires room for terminating null to be included.


arutk:
This change effectively affects only the case where source string is short and not NULL terminated in the first slen bytes. Implicit assumption is that we can safely call strlcpy(dst, src, slen+1) since the source string fits into destination even with terminating NULL. 

For strlcpy to work correctly input string must be NULL terminated. In Linux kernel adapter we assume that strings are NULL terminated. If string is NULL terminated and caller uses this API correctly, then this change has no effect. 

So the conclusion is that this change only handles invalid use case.

PR #126, opened 2019-09-25
Don't generate core name when adding core to corepool.

### LINKED ISSUE #124: casadm --script --add-core --try-add fails if there is no cache with cache_id
casadm --script --add-core --try-add --cache-id 1 --core-device /dev/sdX should work regardless of (non)existing cache with cache_id

mmichal10:
When core frome corepool is added to cache, it's old name is loaded and cores
in corepool cannot be referenced by name anyway so new name is not needed.

Fixes #124 

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -534,8 +534,7 @@ int cache_mngt_prepare_core_cfg(struct ocf_mngt_core_config *cfg,
 		struct kcas_insert_core *cmd_info)
 {
 	struct block_device *bdev;
-	static char core_name[OCF_CORE_NAME_SIZE];
-	struct cache_priv *cache_priv;
+	char core_name[OCF_CORE_NAME_SIZE] = {};
 	ocf_cache_t cache;
 	uint16_t core_id;
 	int result;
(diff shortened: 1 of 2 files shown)


modules/cas_cache/layer_cache_management.c
```diff
@@ -534,8 +534,7 @@ int cache_mngt_prepare_core_cfg(struct ocf_mngt_core_config *cfg,
 		struct kcas_insert_core *cmd_info)
 {
 	struct block_device *bdev;
-	static char core_name[OCF_CORE_NAME_SIZE];
-	struct cache_priv *cache_priv;
+	static char core_name[OCF_CORE_NAME_SIZE] = {};
```
robertbaldyga: This shouldn't be static. It looks like it has been made static by mistake.

modules/cas_cache/layer_cache_management.c
```diff
@@ -546,17 +545,19 @@ int cache_mngt_prepare_core_cfg(struct ocf_mngt_core_config *cfg,
 	if (cmd_info->core_id == OCF_CORE_MAX) {
 		result = mngt_get_cache_by_id(cas_ctx, cmd_info->cache_id,
 				&cache);
-		if (result)
+		if (result && result != -OCF_ERR_CACHE_NOT_EXIST) {
 			return result;
+		} else if (result != -OCF_ERR_CACHE_NOT_EXIST) {
+			struct cache_priv *cache_priv;
+			cache_priv = ocf_cache_get_priv(cache);
```
robertbaldyga: So we are accessing `cache` despite getting it failed? It doesn't look good.

PR #145, opened 2019-10-01
casadm: Add message for error code OCF_ERR_CORE_NOT_EXIST

### LINKED ISSUE #143: Unknown error when removing non-existing core
Reproduction:
1. start cache
2. add core
3. try removing core with different id than added

Expected:
Info that core does not exist

Result:
Error while removing core device X from cache instance 1
Unknown error 1000015

robertbaldyga:
Fixes #143 


arutk:
Already fixed in another previous patchset

PR #156, opened 2019-10-18
env: implement vmalloc with GFP flags

arutk:

### DIFF
--- a/modules/cas_cache/ocf_env.h
+++ b/modules/cas_cache/ocf_env.h
@@ -40,14 +40,24 @@ static inline void env_free(const void *ptr)
 	kfree(ptr);
 }
 
+static inline void *env_vmalloc_flags(size_t size, int flags)
+{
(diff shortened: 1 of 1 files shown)


modules/cas_cache/ocf_env.h
```diff
@@ -40,14 +40,24 @@ static inline void env_free(const void *ptr)
 	kfree(ptr);
 }
 
+static inline void *env_vmalloc_flags(size_t size, int flags)
+{
+	return __vmalloc(size, flags | __GFP_HIGHMEM, PAGE_KERNEL);
```
robertbaldyga: If any kind of kernel flags are going to be passed from OCF code, appropriate ENV_ macros should be provided.

arutk: flags are already defined ocf_env.h, line 19

PR #159, opened 2019-10-21
Update Readme - installation and running tests

robertbaldyga:


README.md
```diff
@@ -21,17 +21,59 @@ https://opensource.org/licenses/BSD-3-Clause for full license texts).
...
 
-The complete documentation for Open CAS Linux is available in the Open CAS Linux Administration Guide
-available [here.](https://open-cas.github.io/guide_introduction.html)
+1. Go to test directory `cd test/functional`.
+1. Install dependencies with command `pip3 install -r test-framework/requirements.txt`.
+1. Create DUT config. See example [here](test/functional/config/example_dut_config.yml).  
+    a) Set disks params. You need at least two disks, of which at least one is an SSD drive.  
+    b) For remote execution uncomment and set the `ip`, `user` and `password` fields.  
+    c) For local execution just leave these fields commented.
+1. Run tests using command `pytest-3 --dut-config=<CONFIG>` where `<CONFIG>` is path to your config file, for example `pytest-3 --dut-config="config/dut_config.yml"`.
```
mmichal10: Wrong enumeration.

robertbaldyga: This is not a bug, this is a feature. This way you can tell Markdown to use autonumbering, so that if you add new item in the middle (or at the very beginning) of the list you don't need to manually update numbers of all the following items.

PR #162, opened 2019-10-22
Use "require_disk" mark

robertbaldyga:


test/functional/tests/conftest.py
```diff
@@ -81,6 +82,9 @@ def prepare_and_cleanup(request):
         test_wrapper.cleanup()
 
 
+def pytest_configure(config):
+    TestRun.configure(config)
+
```
Deixx: pep8 expected 2 blank lines

test/functional/tests/conftest.py
```diff
@@ -64,15 +64,16 @@ def prepare_and_cleanup(request):
...
+    TestRun.setup(dut_config)
 
     TestRun.plugins['opencas'] = {'already_updated': False}
 
-    TestRun.LOGGER.info(f"**********Test {request.node.name} started!**********")
-    yield
+    TestRun.LOGGER.info(f"**********Test {item.name} started!**********")
+
 
+def pytest_runtest_teardown(item, nextitem):
```
Deixx: I assume item and nextitem are required for this to be a python hook? Because I don't like unused variables :D

test/functional/tests/conftest.py
```diff
@@ -35,8 +35,7 @@ def get_pytest_options(request):
     pytest_options["force_reinstall"] = request.config.getoption("--force-reinstall")
 
 
-@pytest.fixture()
-def prepare_and_cleanup(request):
+def pytest_runtest_setup(item):
     """
     This fixture returns the dictionary, which contains DUT ip, IPMI, spider, list of disks.
```
Deixx: description needs to be updated

test/functional/tests/io_class/test_io_class_directory.py
```diff
@@ -17,11 +17,10 @@
 from .io_class_common import *
 
 
+@pytest.mark.require_disk("cache", [DiskType.optane, DiskType.nand])
+@pytest.mark.require_disk("core", [DiskType.sata, DiskType.hdd])
```
Deixx: this is a limitation for cores - we usually just want cores to be "lower" type than caches, so if we have optanes as caches, we don't want to exclude nands as possible cores, and that's what happened here

test/functional/tests/conftest.py
robertbaldyga: Yes, they are.

test/functional/tests/conftest.py
robertbaldyga: Done.

test/functional/tests/io_class/test_io_class_directory.py
robertbaldyga: Done.

test/functional/tests/conftest.py
robertbaldyga: Done.

test/functional/tests/cli/test_cli_start_stop.py
```diff
@@ -7,18 +7,15 @@
 import pytest
 from api.cas import casadm, casadm_parser
 from core.test_run import TestRun
-from storage_devices.disk import DiskType
+from storage_devices.disk import DiskType, DiskTypeSet
 from test_utils.size import Unit, Size
 
 
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane]))
```
micrakow: Are cli tests related to the Optanes? I think that all types or at least optane & nand should work here, don't think there's a reason to limit it to Optane only

micrakow: Are cli tests related to the Optanes? I think that all types or at least optane & nand should work here, don't think there's a reason to limit it to Optane only

robertbaldyga: It was already there, I didn't touch this. However I think you can expect another patch fixing this problem in the near future.

PR #164, opened 2019-10-23
Enable logging mechanism and example of logger usage in one test

katlapinka:
Logging mechanism enabled in one example test.
Creating logs enabled in conftest.
Gathering additional logs at the end of test.
Created cleanup method (teardown) in conftest, which is executed at the end of each test, even if test  was interrupted/failed.

### DIFF
--- a/test/functional/tests/conftest.py
+++ b/test/functional/tests/conftest.py
@@ -7,16 +7,20 @@ import pytest
 import os
 import sys
 import yaml
+import traceback
 from IPy import IP
+
+from connection.ssh_executor import SshExecutor
+
 sys.path.append(os.path.join(os.path.dirname(__file__), "../test-framework"))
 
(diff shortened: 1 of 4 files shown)


test/functional/tests/conftest.py
```diff
@@ -35,6 +40,35 @@ def get_pytest_options(request):
...
+def pytest_runtest_teardown():
+    """
+    This method is executed always in the end of each test, even if it fails or raises exception in
+    prepare stage.
+    """
+    TestRun.LOGGER.end_all_groups()
+
+    with TestRun.LOGGER.step("Cleanup after test"):
+        try:
+            type_e = type(TestRun.executor) is SshExecutor
```
micrakow: how about renaming type_e to smth like 'ssh_e' or similar?

katlapinka:
retest this please

PR #166, opened 2019-10-24
Fix startup

imjfckm:
### Now
Current startup procedure works on an assumption that we will
deal with asynchronously appearing devices in asynchronous way
(udev rules) and synchronous events in the system (systemd units)
won't interfere. If we would break anything (mounts) we would just
take those units and restart them. This tactic was working as long
as resetting systemd units took reasonable time.

### Why it failed
As hackish as it sounds it worked in all systems that the software
has been validated on. Unfortunately it stopped working because
of *.mount units taking MUCH longer time to restart even on
mainstream OSes, so it's time to change.

### What is this PR doing
This change implements open-cas systemd service which will wait
synchronously with systemd bootup process for all required Open CAS
devices to start. If they don't we fail the boot process just as
failing mounts would. We also make sure that this process takes place
before any mounts (aside from root FS and other critical FS's) are
even attempted. Now opencas-mount-utility can be discarded.

### How to use
To override this behaviour on per-core basis you can specify
lazy_startup=true option in opencas.conf.

### What next
This PR only deals with synchronously working services that play nice with systemd. There is also the case of device mapper, LVM and similar which don't respect systemd targets (and why should they?). I do have some idea on how to fix it which I'll be testing over next few days and if I succeed I'll submit another PR.


utils/opencas.py
```diff
@@ -689,3 +736,71 @@ def stop(flush):
...
+            if (
+                core.params.get("lazy_load", None)
+                and core.params["lazy_load"] == "true"
+            ):
+                continue
+
+            runtime_state = runtime_core_state.get(core.core_id, None)
+            if not runtime_state or runtime_state["status"] != "Active":
+                not_initialized.append(core)
+
```
mmichal10: Condition to `continiue` is true for all `target_core_state` elements.

utils/opencas.py
```diff
@@ -689,3 +736,71 @@ def stop(flush):
...
+        config = cas_config.from_file(
+            cas_config.default_location, allow_incomplete=True
+        )
+    except Exception as e:
+        raise Exception("Unable to load opencas config. Reason: {0}".format(str(e)))
+
+    stop_time = time.time() + int(timeout)
+
+    not_initialized = None
+    target_core_state = config.get_startup_cores()
```
mmichal10: This variable and ` config.get_startup_cores()` are unneeded.

utils/opencas.py
```diff
@@ -689,3 +736,71 @@ def stop(flush):
...
+            if core_pool:
+                devices["core_pool"].append(core)
+            else:
+                core.update({"cache_id": prev_cache_id})
+                devices["cores"].update({int(device["id"]): core})
+
+    return devices
+
+
+def wait_for_startup(timeout=300, interval=4, wait_for_lazy=False):
```
mmichal10: `wait_for_lazy` is never used.

imjfckm: True that, removed

utils/opencas.py
imjfckm: I think it is used, we need it to check if we've arrived at desired configuration

utils/opencas.py
imjfckm: Close, but no cigar! Condition is always false! Removed :)

utils/opencas.py
```diff
@@ -353,7 +380,20 @@ def check_recursive(self):
                 raise ValueError('Recursive configuration detected')
 
         def to_line(self):
-            return '{0}\t{1}\t{2}\n'.format(self.cache_id, self.core_id, self.device)
+            ret = "{0}\t{1}\t{2}".format(self.cache_id, self.core_id, self.device)
+            if len(self.params) > 0:
```
mmichal10: Would it be possible to remove this ` if ` statement?

utils/opencas.py
```diff
@@ -689,3 +736,65 @@ def stop(flush):
...
+
+    devices = {"core_pool": [], "caches": {}, "cores": {}}
+
+    core_pool = False
+    prev_cache_id = -1
+
+    for device in device_list:
+        if device["type"] == "core pool":
+            core_pool = True
+            continue
```
mmichal10: I believe that adding newline would increase readability a little.

utils/opencas.py
```diff
@@ -689,3 +736,65 @@ def stop(flush):
...
+            if core_pool:
+                devices["core_pool"].append(core)
+            else:
+                core.update({"cache_id": prev_cache_id})
+                devices["cores"].update({int(device["id"]): core})
+
+    return devices
+
+
+def wait_for_startup(timeout=300, interval=4):
```
mmichal10: cli default interval value is 5, so maybe we should be more consistent about it?

utils/opencas.py
```diff
@@ -689,3 +736,65 @@ def stop(flush):
...
+                    }
+                }
+            )
+        elif device["type"] == "core":
+            core = {"device": device["disk"], "status": device["status"]}
+            if core_pool:
+                devices["core_pool"].append(core)
+            else:
+                core.update({"cache_id": prev_cache_id})
+                devices["cores"].update({int(device["id"]): core})
```
mmichal10: Core id is used as a key in `devices["cores"]` dict. And if it is used more than one, each new object will override value of the old one.

utils/opencas.py
```diff
@@ -353,7 +380,20 @@ def check_recursive(self):
                 raise ValueError('Recursive configuration detected')
 
         def to_line(self):
-            return '{0}\t{1}\t{2}\n'.format(self.cache_id, self.core_id, self.device)
+            ret = "{0}\t{1}\t{2}".format(self.cache_id, self.core_id, self.device)
+            if len(self.params) > 0:
+                i = 0
+                for param, value in self.params.items():
+                    if i > 0:
```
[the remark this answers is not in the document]
imjfckm: Done

utils/opencas.py
imjfckm: Done

utils/opencas.py
imjfckm: Done

utils/opencas.py
imjfckm: Done

utils/opencas.py
imjfckm: Nice catch. Changed it up to be keyed with (cache_id, core_id) pair. Also, added tests.

test/utils_tests/opencas-py-tests/test_helper_functions_01.py
```diff
@@ -0,0 +1,583 @@
+#
+# Copyright(c) 2012-2019 Intel Corporation
```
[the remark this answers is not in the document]
imjfckm: Fixed

PR #172, opened 2019-10-30
tests: Update installer

robertbaldyga:


test/functional/api/cas/installer.py
```diff
@@ -9,68 +9,49 @@
...
+    TestRun.LOGGER.info("Building Open CAS")
     output = TestRun.executor.run(
-        f"cd {opencas_repo_name} && "
-        "git submodule update --init --recursive && "
+        f"cd {TestRun.plugins['opencas']['working_dir']} && "
         "./configure && "
         "make -j")
     if output.exit_code != 0:
         raise Exception(
             f"Make command executed with nonzero status: {output.stdout}\n{output.stderr}")
```
katlapinka: please use TestRun.LOGGER.exception instead of raising exception, so output could be visible in test log.

robertbaldyga:
> Warnings during make -j causes exceptions, please do some workaround for it

We compile code with `-Werror` flag so every warning automatically becomes error which should case an exception, thus I think current behavior is correct.

test/functional/api/cas/installer.py
robertbaldyga: Done.

PR #173, opened 2019-10-31
Add cache config and fix for CAS modules names

katlapinka:


test/functional/api/cas/cache_config.py
```diff
@@ -110,5 +111,22 @@ def default_seq_cut_off_params():
...
+                 eviction_policy=EvictionPolicy.DEFAULT,
+                 metadata_mode=MetadataMode.normal):
+        self.cache_line_size = cache_line_size
+        self.cache_mode = cache_mode
+        self.cleaning_policy = cleaning_policy
+        self.eviction_policy = eviction_policy
+        self.metadata_mode = metadata_mode
+
+
+def validate_cache_params(cache, cache_config: CacheConfig):
```
robertbaldyga: I think it would be clearer if we had implemented `__eq__()` in `CacheConfig`. Then we could just do:
```python
assert config == cache.get_cache_config()
```

PR #181, opened 2019-11-06
tests: Update example using new utils

robertbaldyga:
Needs: https://github.com/Open-CAS/test-framework/pull/62

### DIFF
--- a/test/functional/tests/example/example_test.py
+++ b/test/functional/tests/example/example_test.py
@@ -19,37 +19,69 @@ def setup_module():
 
 @pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
 def test_create_example_partitions():
-    TestRun.LOGGER.info("Test run")
-    TestRun.LOGGER.info(f"DUT info: {TestRun.dut}")
-    test_disk = TestRun.disks['cache']
-    part_sizes = []
-    for i in range(1, 6):
-        part_sizes.append(Size(10 * i + 100, Unit.MebiByte))
-    test_disk.create_partitions(part_sizes)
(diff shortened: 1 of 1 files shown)


test/functional/tests/example/example_test.py
```diff
@@ -19,37 +19,67 @@ def setup_module():
...
+        file1 = File.create_file("example_file")
+        file1.write("Test file\ncontent line\ncontent")
+    with TestRun.step("Read file content"):
+        content_before_change = file1.read()
+        TestRun.LOGGER.info(f"File content: {content_before_change}")
+    with TestRun.step("Replace single line in file"):
+        fs_utils.replace_in_lines(file1, 'content line', 'replaced line')
+    with TestRun.step("Read file content and check if it changed"):
+        content_after_change = file1.read()
+        assert content_before_change != content_after_change
```
katlapinka: please do not use assert but TestRun.fail()

test/functional/tests/example/example_test.py
```diff
@@ -19,37 +19,67 @@ def setup_module():
...
+
+    with TestRun.step("Make copy of the file and check if md5 sum matches"):
+        file2 = file1.copy('/tmp', force=True)
+        assert file1.md5sum() == file2.md5sum()
```
katlapinka: as before - use TestRun.fail()

test/functional/tests/example/example_test.py
robertbaldyga: Done.

test/functional/tests/example/example_test.py
robertbaldyga: Done.

PR #182, opened 2019-11-07
Move test log creation to test-framework

katlapinka:


test/functional/tests/conftest.py
```diff
@@ -48,7 +48,10 @@ def pytest_runtest_setup(item):
     TestRun.prepare(item)
 
     test_name = item.name.split('[')[0]
-    TestRun.LOGGER = create_log(item.config.getoption('--log-path'), test_name)
+    test_params = []
+    for name, value in item.callspec.params.items():
+        test_params.append(f"{name}-{value}")
+    TestRun.LOGGER = create_log(item.config.getoption('--log-path'), test_name, test_params)
```
robertbaldyga: We can do all of this internally in test-framework. I believe other users of test-framework will also want to have test name & options printed.

katlapinka: Done

Issue #165, opened 2019-10-24
Kernel oops on modprobe cas_cache


ndcioban:
I'm getting the following kernel oops when loading cas_cache:

[   76.666811] BUG: unable to handle kernel NULL pointer dereference at 0000000000000000
[   76.676546] PGD 0 P4D 0
[   76.679676] Oops: 0002 [#1] PREEMPT SMP [   76.706648] Workqueue: events _cas_rpool_pre_alloc_do [cas_cache]
[   76.714251] RIP: 0010:_cas_rpool_pre_alloc_do+0x54/0xa0 [cas_cache]
[   76.722043] Code: 48 c1 e3 04 49 03 5 GS:ffff88a6e4300000(0000) knlGS:0000000000000000
[   76.806438] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[   76.813597] CR2: 0000000000000000 CR3: 000000002a012002 CR4: 00000000003606e0
[   76.822470] DR0: 0000000000000000 DR1: 0000000000000000 DR2: 0000000000000000
[   76.831355] DR3: 0000000000000000 DR6: 00000000fffe0ff0 DR7: 0000000000000400
[   76.840252] Call Trace:
[   76.843316]  process_one_work+0x1d5/0x316
[   76.848329]  ? create_worker+0x1dc/0x1dc
[   76.853214]  worker_thread+0x1cd/0x2bf
[   76.857848]  ? create_worker+0x1dc/0x1dc
[   76.862698]  kthread+0x14f/0x157

I'm using open-cas-linux on v4.19.80 kernel and configured the cache device to be an NVMe optane drive. 

My open-cas-linux sha1 is 06b3e69c0f7b

nciobanu:
Some more debugging - looks like it is somehow related to having only some logical cores online but not all. If some cores are disabled in OS (sysfs) then this bug occurs. If  I force enable all the cores then cas_cache loads fine. 

I put some printks to understand what's going on:

`diff --git a/modules/cas_cache/utils/utils_rpool.c b/modules/cas_cache/utils/utils_rpool.c
index 560579c5f278..7d4381b859ab 100644
--- a/modules/cas_cache/utils/utils_rpool.c
+++ b/modules/cas_cache/utils/utils_rpool.c
@@ -65,11 +65,18 @@ void _cas_rpool_pre_alloc_do(struct work_struct *ws)
        void *entry;
        int i, cpu;

[...]

PR #188, opened 2019-11-15
Fix page fault in rpool allocation to offline CPUs

nciobanu:
In cases where HT is enabled in BIOS but disabled in OS
we run into a situation where we are allocating resources
and attempting to schedule work on offline cores which causes
a page fault and resets the machine.

This change checks the number of possible CPUs and checks
that each is online before allocating rpools. Fixes issue #165

Signed-off-by: Nathan Ciobanu <nathan.d.ciobanu@intel.com>

### DIFF
--- a/modules/cas_cache/utils/utils_rpool.c
+++ b/modules/cas_cache/utils/utils_rpool.c
@@ -105,7 +105,7 @@ int _cas_rpool_pre_alloc_schedule(int cpu,
 void cas_rpool_destroy(struct cas_reserve_pool *rpool_master,
 		cas_rpool_del rpool_del, void *allocator_ctx)
 {
-	int i, cpu_no = num_online_cpus();
+	int i, max_cpus = num_possible_cpus();
 	struct _cas_reserve_pool_per_cpu *current_rpool = NULL;
 	struct list_head *item = NULL, *next = NULL;
 	void *entry;
@@ -120,7 +120,10 @@ void cas_rpool_destroy(struct cas_reserve_pool *rpool_master,
 		return;
 	}
 
-	for (i = 0; i < cpu_no; i++) {
+	for (i = 0; i < max_cpus; i++) {
+		if (!cpu_online(i))
+			continue;
+
 		current_rpool = &rpool_master->rpools[i];
 
 		CAS_DEBUG_PARAM("Destroyed reserve poll [%s] for cpu %d",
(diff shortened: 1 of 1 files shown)


mmichal10:
Ok to test

nciobanu:
I've found several more places where we don't check if cpu is online and attempt to create threads so I need to update this patch. Let's not merge it just yet

modules/cas_cache/utils/utils_rpool.c
```diff
@@ -120,7 +120,10 @@ void cas_rpool_destroy(struct cas_reserve_pool *rpool_master,
 		return;
 	}
 
-	for (i = 0; i < cpu_no; i++) {
+	for (i = 0; i < max_cpus; i++) {
```
robertbaldyga: If some CPU that was online on CAS load will go offline during CAS operation, here we will have memory leak.
If some CPU that was offline on CAS load will go online during CAS operation, here we will be attempting to deallocate memory we didn't allocated.

robertbaldyga:
> I've found several more places where we don't check if cpu is online and attempt to create threads so I need to update this patch. Let's not merge it just yet

@nciobanu Are you planning to update this pull request? Otherwise I'm going to close it.

nciobanu:
I'm planning on sending an updated patch

modules/cas_cache/utils/utils_rpool.c
```diff
@@ -177,7 +180,10 @@ struct cas_reserve_pool *cas_rpool_create(uint32_t limit, char *name,
 	info.rpool_new = rpool_new;
 	info.allocator_ctx = allocator_ctx;
 
-	for (i = 0; i < cpu_no; i++) {
+	for (i = 0; i < max_cpus; i++) {
+		if (!cpu_online(i))
```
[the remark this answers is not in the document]
nciobanu: I think this case is covered by the fact that I'm kzalloc * max_cpus. If anything we are allocating more memory than possibly needed but it should cover this case. 

An argument can be made that we may allocate memory unnecessarily in most situations but since this is happening only on initialization it would probably require a more complicated mechanism to handle that case. 

What would you recommend?

modules/cas_cache/utils/utils_rpool.c
nciobanu: > If some CPU that was online on CAS load will go offline during CAS operation, here we will have memory leak.
I think this case will be fixed if I remove the two lines below. 

> If some CPU that was offline on CAS load will go online during CAS operation, here we will be attempting to deallocate memory we didn't allocated.
I think this is safe because we allocate memory for num_possible_cpus(0 in cas_rpool_create().

nciobanu: Also please advise if my code flow assumptions are wrong since I'm not deeply familiar with the project overall.

robertbaldyga:
I'm closing due to lack of activity.

PR #194, opened 2019-11-22
Extend dut config file

mmichal10:


test/functional/tests/conftest.py
```diff
@@ -181,6 +181,11 @@ def base_prepare(item):
             except Exception:
                 pass  # TODO: Reboot DUT if test is executed remotely
 
+        for disk in TestRun.dut.disks:
+            disk.umount_all_partitions()
+            if not disk_utils.create_partition_table(disk, disk_utils.PartitionTable.gpt):
+                raise Exception(f"Failed to remove partitions from {disk}")
```
katlapinka: use TestRun.exception instead of raising

PR #195, opened 2019-11-23
tests: Improve flush handling during cache mode change

rafalste:


mmichal10:
add to whitelist

PR #200, opened 2019-11-26
tests: Handle pytest test state properly

robertbaldyga:
- Log information about test fail in case of exception.
- Skip teardown in case of test skip (setup is skipped by default).


robertbaldyga:
retest this please

PR #211, opened 2019-11-29
Add FI test: try add twice the same device as core 

Ostrokrzew:
Check that IntelCAS does not accepts twice the same device as core:
/*1*/ "Start cache",
/*2*/ "Add Core device to cache",
/*3*/ "Try add as second core device which was attached previously",
/*4*/ "Remove IntelCAS device"

Signed-off-by: Slawomir_Jankowski <slawomir.jankowski@intel.com>

### DIFF
--- a/test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
+++ b/test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
@@ -184,3 +184,46 @@ def test_add_occupied_core(cache_mode):
 
     with TestRun.step("Stop all caches."):
         casadm.stop_all_caches()
+
+
+@pytest.mark.parametrize("cache_mode", CacheMode)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_add_twice_same_core(cache_mode):
+    """
+        title: Test if OpenCAS not accepts twice the same device as core.
+        description: |
+          False positive test of the ability to add core twice to OpenCAS.
+        pass_criteria:
+          - No system crash.
+          - The same core device cannot be used twice in CAS.
+    """
+    with TestRun.step("Prepare cache and core."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(2, Unit.GibiByte)])
+        cache_dev = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_dev.create_partitions([Size(1, Unit.GibiByte)])
+        core_dev = core_dev.partitions[0]
+
+    with TestRun.step("Start OpenCAS"):
+        cache = casadm.start_cache(cache_dev, cache_mode, force=True)
+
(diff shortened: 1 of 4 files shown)


test/functional/tests/stress/test_stress_without_io.py
```diff
@@ -180,7 +178,12 @@ def test_stress_reload_cache(cache_mode):
                 TestRun.fail(f"Expected cores count: 1; Actual cores count: {cores_count}.")
 
     with TestRun.step("Check md5 of test file."):
-        check_files(core, size_before_load, permissions_before_load, md5_before_load)
+        core.mount(mount_point)
+        file_after = file.get_properties()
+        if file_after != file:
```
robertbaldyga: So here we compare file with it's own properties?
The first is object of `File`, the second is object of `FileProperties`. Doesn't it throw exception or something?

test/functional/tests/stress/test_stress_without_io.py
```diff
@@ -221,7 +223,12 @@ def test_stress_add_remove_core(cache_mode):
...
```
robertbaldyga: Ditto.

test/functional/tests/stress/test_stress_without_io.py
```diff
@@ -270,47 +276,25 @@ def test_stress_reload_module(cache_mode):
                 TestRun.fail("Cache configuration is different than before reloading modules.")
 
     with TestRun.step("Check md5 of test file."):
-        check_files(core, size_before_load, permissions_before_load, md5_before_load)
+        core.mount(mount_point)
+        file_after = file.get_properties()
+        if file_after != file:
```
robertbaldyga: Ditto.

test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
```diff
@@ -0,0 +1,231 @@
...
+        if caches_count != 1:
+            TestRun.fail(f"Expected caches count: 1 Actual caches count: {caches_count}.")
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != 0:
+            TestRun.fail(f"Expected cores count: 0; Actual cores count: {cores_count}.")
+
+    with TestRun.step("Check md5 of test file."):
+        file1 = file.get_properties()
+        file2 = copied_file.get_properties()
+        if file1 != file2:
```
robertbaldyga: Names `file1` and `file2` are misleading in this context.
Why don't we just do this?
```python
if file1.get_properties() != file2.get_properties():
```

test/functional/tests/stress/test_stress_without_io.py
Ostrokrzew: No, because object of File (which inherits after FsItem) has the same properties to compare with FileProperties' object. But You're right, for clarity and explicitness there should be compared two objects of the same class. I'll do that.

test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
Ostrokrzew: You were right, this could be confusing, so I changed it.

PR #212, opened 2019-11-29
utils: lengthen open-cas service timeout

liuhongtong:
Signed-off-by: liuhongtong <hongtongliu@126.com>

### DIFF
--- a/utils/open-cas.service
+++ b/utils/open-cas.service
@@ -15,7 +15,7 @@ OnFailureJobMode=isolate
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/sbin/casctl settle --timeout 150 --interval 5
+ExecStart=/sbin/casctl settle --timeout 1780 --interval 5
 TimeoutStartSec=30min
 
 [Install]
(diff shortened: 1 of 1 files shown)


robertbaldyga:
ok to test

imjfckm:
Thanks for the PR, @liuhongtong !
Why exactly do you want this timeout to be longer? Is there any specific issue you're encountering?
Also, timeout in systemd service file is just a safeguard in case something goes really wrong - the real timeout you should be looking at is a line above in `ExecStart=/sbin/casctl settle --timeout 150 --interval 5`. This is the time after which te service will fail if devices won't start.

liuhongtong:
@imjfckm Thanks for your attention.
There are at least four NVMe SSDs in my system, maybe more in the future. Init cache devices needs a long time, that leads to  open-cas service starting failure and os entering emergency mode.

imjfckm:
Sounds reasonable to me! I do have some doubts still. Does the patch fix it for you? Without changing the timeout in `ExecStart` ?

liuhongtong:
@imjfckm Thanks. I will retest this issue.

imjfckm:
@liuhongtong any update on this?

PR #224, opened 2019-12-11
Add API to validate CAS CLI messages

katlapinka:


test/functional/api/cas/cli_messages.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+import re
+
+from core.test_run import TestRun
+from test_utils.output import Output
+
+
+load_inactive_core_missing = [
+    r"WARNING: Can not resolve path to core \d from cache \d\. By-id path will be shown for that "
```
Deixx: We need '\d+' instead of '\d' everywhere (IDs can have more than one digit)

katlapinka: done

PR #218, opened 2019-12-06
Add new statistics API

rafalste:

PR #225, opened 2019-12-11
Adapt CAS API to handle disk plugging and fix for setting clean…

katlapinka:
…ing/cutoff params

### DIFF
--- a/test/functional/api/cas/cache.py
+++ b/test/functional/api/cas/cache.py
@@ -160,15 +160,22 @@ class Cache:
 
     def set_params_acp(self, acp_params: FlushParametersAcp):
         return casadm.set_param_cleaning_acp(self.cache_id,
-                                             acp_params.wake_up_time.total_milliseconds(),
-                                             acp_params.flush_max_buffers)
+                                             int(acp_params.wake_up_time.total_milliseconds())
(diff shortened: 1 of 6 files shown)


test/functional/api/cas/cache.py
```diff
@@ -65,7 +65,7 @@ def get_occupancy(self):
         return self.get_cache_statistics()["occupancy"]
 
     def get_status(self):
-        status = self.get_cache_statistics()["status"].replace(' ', '_')
+        status = self.get_cache_statistics()["status"].replace(' ', '_').lower()
```
rafalste: If PR #218 gets merged this line should look like this:
`status = self.get_statistics().config_stats.status.replace(' ', '_').lower()`

test/functional/tests/conftest.py
```diff
@@ -117,7 +118,10 @@ def pytest_runtest_teardown():
                     TestRun.executor.wait_for_connection()
                 Udev.enable()
                 unmount_cas_devices()
+                casadm.remove_all_detached_cores()
                 casadm.stop_all_caches()
+                from api.cas import init_config
```
imjfckm: Why import here and not at the top of the file?

PR #230, opened 2019-12-12
Cache configuration load - every mode - 1 cache 1 or 4 cores

Ostrokrzew:
Iterate over all of valid combinations of the start options e.g.: eviction policy LRU/LMP, metadata variants MAX/MIX/MIN (overall 12 iterations), cleaning policy ALRU/NOP/ACP:
1. Start cache in write-back mode with configuration.
2. Configure cleaning policy.
3. Add cores.
4. Run Vdbench on created CAS device (eviction must occur during this run).
5. Stop cache.
6. Start cache again, load the configuration.
7. Check if cache configuration was loaded properly.
8. Run Vdbench again.
9. Stop cache.

Iterate over all of valid combinations of the start options e.g.: eviction policy LRU/LMP, metadata variants MAX/MIX/MIN (overall 12 iterations), cleaning policy ALRU/NOP:
1. Start cache in write-back mode with configuration.
2. Configure cleaning policy.
3. Add core.
4. Run Vdbench on created CAS device (eviction must occur during this run).
5. Stop cache.
6. Start cache again, without specifying eviction policy, metadata variant and cleaning policy.
7. Check if cache configuration was loaded properly.
8. Run Vdbench again.
9. Stop cache.

But there's only 1 eviction policy and no metadata variants in OpenCAS, so only cache line size and  cleaning policy are changing.

Signed-off-by: Ostrokrzew <ostrokrzew@protonmail.com>

PR #242, opened 2019-12-18
Fix minor issues in API

rafalste:
- Fix typo in 'get ACP parameters' function
- Fix return type of Time class (from float to int)
- Add string representation for cleaning policy

### DIFF
--- a/test/functional/api/cas/casadm_parser.py
+++ b/test/functional/api/cas/casadm_parser.py
@@ -190,7 +190,7 @@ def get_flush_parameters_alru(cache_id: int):
 
 def get_flush_parameters_acp(cache_id: int):
     casadm_output = casadm.get_param_cleaning_acp(cache_id,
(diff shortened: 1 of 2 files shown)


test/functional/api/cas/casadm_parser.py
```diff
@@ -190,7 +190,7 @@ def get_flush_parameters_alru(cache_id: int):
 
 def get_flush_parameters_acp(cache_id: int):
     casadm_output = casadm.get_param_cleaning_acp(cache_id,
-                                                  casadm.OutputFormat.csv).stdout.spltlines()
+                                                  casadm.OutputFormat.csv).stdout.splitlines()
```
Ostrokrzew: Also fixed in https://github.com/Open-CAS/open-cas-linux/pull/230 PR and it's not only typo here.

PR #251, opened 2019-12-24
Add stress test for small caches

Ostrokrzew:
Add IO stress tests for small CAS devices with 1 or 4 cores,
for every cache mode and cache line size,
also with every cleaning policy for WB and WO modes.

Signed-off-by: Ostrokrzew <slawomir.jankowski@intel.com>

### DIFF
--- a/test/functional/tests/stress/test_stress_io_small_device.py
+++ b/test/functional/tests/stress/test_stress_io_small_device.py
@@ -1,0 +1,107 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+from datetime import timedelta
+
+from api.cas import casadm, casadm_parser
+from api.cas.cache_config import CacheLineSize, CacheMode, CleaningPolicy
+from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeLowerThan, DiskTypeSet
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import CpusAllowedPolicy, IoEngine, ReadWrite
+from test_utils.size import Size, Unit
+
+stress_time = timedelta(minutes=30)
+
+
+@pytest.mark.parametrize("cores_number", [1, 4])
+@pytest.mark.parametrize("cache_config", [(CacheMode.WT, None),
+                                          (CacheMode.WA, None),
+                                          (CacheMode.PT, None),
+                                          (CacheMode.WB, CleaningPolicy.acp),
+                                          (CacheMode.WB, CleaningPolicy.alru),
+                                          (CacheMode.WB, CleaningPolicy.nop),
+                                          (CacheMode.WO, CleaningPolicy.acp),
+                                          (CacheMode.WO, CleaningPolicy.alru),
+                                          (CacheMode.WO, CleaningPolicy.nop)])
+@pytest.mark.parametrize("cache_line_size", CacheLineSize)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stress_small_cas_device(cache_line_size, cores_number, cache_config):
+    """
+        title: Stress test for verifying data on small CAS devices.
+        description: |
+          Validate the ability of CAS to handle many iops when device is small
+          using different cache modes, cache line sizes and core numbers.
+        pass_criteria:
+          - No system crash.
+          - Md5 sums of core device and exported object are equal.
+    """
+    cache_mode, cleaning_policy = cache_config
+
+    with TestRun.step(f"Prepare 1 cache and {cores_number} core devices."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(100, Unit.MebiByte)])
+        cache_part = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_sizes = [Size(200, Unit.MebiByte)] * cores_number
+        core_dev.create_partitions(core_sizes)
+
+    with TestRun.step(f"Start cache with {cores_number} cores."):
+        cache = casadm.start_cache(cache_part, cache_mode, cache_line_size, force=True)
+        cores = []
+        for i in range(cores_number):
+            cores.append(cache.add_core(core_dev.partitions[i]))
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 1:
+            TestRun.fail(f"Expected caches count: 1; Actual caches count: {caches_count}.")
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != cores_number:
+            TestRun.fail(
+                f"Expected cores count: {cores_number}; Actual cores count: {cores_count}.")
+
+    if cleaning_policy is not None:
+        with TestRun.step("Set cleaning policy."):
+            cache.set_cleaning_policy(cleaning_policy)
+
+    with TestRun.step(f"Stress cache for {int(stress_time.total_seconds() / 60)} minutes."):
+        fio = (Fio().create_command()
+               .io_engine(IoEngine.libaio)
+               .io_depth(128)
+               .direct()
+               .time_based()
+               .run_time(stress_time)
+               .read_write(ReadWrite.randrw)
+               .block_size(cache_line_size)
+               .num_jobs(cores_number)
+               .cpus_allowed_policy(CpusAllowedPolicy.split))
+        for core in cores:
+            fio.add_job(f"job_{core.core_id}").target(core.system_path)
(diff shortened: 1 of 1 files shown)


test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+from core.test_run import TestRun
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from test_utils.size import Size, Unit
+
+stress_time = timedelta(minutes=30)
+
+
+@pytest.mark.parametrize("cache_mode", [CacheMode.WT, CacheMode.WA, CacheMode.PT])
+@pytest.mark.parametrize("core_number", [1, 4])
```
micrakow: Please change to smth like 'cores_number' or 'cores_count'

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.parametrize("cache_line_size", CacheLineSize)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stress_small_cas_device(cache_mode, cache_line_size, core_number, cleaning_policy=None):
+    """
+        title: Stress test for verifying data on small CAS devices.
+        description: |
+          Validate the ability of the CAS to handle many iops when device is small
```
micrakow: Are iops really being checked in this test or only data consistency is?

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+        pass_criteria:
+          - No system crash.
+          - Md5 sums of core device and cache device are equal.
+    """
+    with TestRun.step(f"Prepare 1 cache and {core_number} core devices."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(100, Unit.MebiByte)])
+        cache_dev = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_size = []
```
micrakow: I think that cores_size would be more descriptive since it is a list

robertbaldyga: I think core_sizes would be more descriptive.

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.mark.parametrize("cleaning_policy", CleaningPolicy)
+@pytest.mark.parametrize("cache_mode", [CacheMode.WB, CacheMode.WO])
+@pytest.mark.parametrize("core_number", [1, 4])
+@pytest.mark.parametrize("cache_line_size", CacheLineSize)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stress_small_cas_device_wb_wo_mode(cache_mode, cache_line_size,
+                                            core_number, cleaning_policy):
+    test_stress_small_cas_device(cache_mode, cache_line_size, core_number, cleaning_policy)
```
robertbaldyga: Calling one test case from another looks weird. How about making it separate function and calling it from both test cases with appropriate params?

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stress_small_cas_device(cache_mode, cache_line_size, core_number, cleaning_policy=None):
+    """
+        title: Stress test for verifying data on small CAS devices.
+        description: |
+          Validate the ability of the CAS to handle many iops when device is small
+          using different cache modes, cache line sizes and core numbers.
+        pass_criteria:
+          - No system crash.
+          - Md5 sums of core device and cache device are equal.
```
Deixx: you are checking core and core device (cache shouldn't have md5 sum equal to core

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != core_number:
+            TestRun.fail(f"Expected cores count: {core_number}; Actual cores count: {cores_count}.")
+
+    if cleaning_policy is not None:
+        with TestRun.step("Set cleaning policy."):
+            cache.set_cleaning_policy(cleaning_policy)
+
+    with TestRun.step(f"Stress cache for {int(stress_time.total_seconds() / 60)} minutes."):
+        for core in cores:
```
Deixx: run single fio with 4 jobs (each job should specify different core as target) - consult me if you need help with that

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,107 @@
...
+            Fio().create_command() \
+                .io_engine(IoEngine.libaio) \
+                .io_depth(128) \
+                .direct() \
+                .time_based() \
+                .run_time(stress_time) \
+                .read_write(ReadWrite.randrw) \
+                .block_size(cache_line_size) \
+                .target(f"{core.system_path}") \
```
Deixx: core.system_path is a string - putting it inside f-string is unnecessary

test/functional/tests/stress/test_stress_io_small_device.py
Deixx: putting steps inside a method will affect the ability to parse test steps properly (and we aim for that)
I wonder if we can make single test and parametrize like ("cache_mode_cleaning_policy", [(WB, [CleaningPolicy]), (WT, None), (WO, [CleaningPolicy]), ...]

robertbaldyga: As long as all the test logic is contained within steps there is no problem with parsing, as we can prepare special test execution mode that skips actual code block within test step. That would produce simple log containing test step titles as a result of simplified test execution, and so having steps in separate function is not a problem (as long as it's not called from test step, but that would be equivalent of having nested test steps which is illegal anyway).

That kind of composite parametrization you proposed is technically doable, but if one day we decide to make some kind of requirement <-> test configuration mapping we're gonna have hard time with that.

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,111 @@
...
+        title: Stress test for verifying data on small CAS devices.
+        description: |
+          Validate the ability of CAS to handle many iops when device is small
+          using different cache modes, cache line sizes and core numbers.
+        pass_criteria:
+          - No system crash.
+          - Md5 sums of core device and exported object are equal.
+    """
+    cache_mode = cache_config[0]
+    cleaning_policy = cache_config[1]
```
robertbaldyga:
```python
cache_mode, cleaning_policy = cache_config
```
-1 to lines count
+100 to pythonicity

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,111 @@
...
+
+    with TestRun.step(f"Prepare 1 cache and {cores_number} core devices."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(100, Unit.MebiByte)])
+        cache_part = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_sizes = []
+        for i in range(cores_number):
+            core_sizes.append(Size(200, Unit.MebiByte))
```
robertbaldyga:
```python
core_sizes = [Size(200, Unit.MebiByte)] * cores_number
```
-2 to lines count
+120 to pythonicity

test/functional/tests/stress/test_stress_io_small_device.py
Ostrokrzew: true, forget that

test/functional/tests/stress/test_stress_io_small_device.py
Ostrokrzew: One more time You reminded me simple techniques.

test/functional/tests/stress/test_stress_io_small_device.py
```diff
@@ -0,0 +1,108 @@
...
+               .io_depth(128)
+               .direct()
+               .time_based()
+               .run_time(stress_time)
+               .read_write(ReadWrite.randrw)
+               .block_size(cache_line_size)
+               .num_jobs(cores_number)
+               .cpus_allowed_policy(CpusAllowedPolicy.split))
+        for core in cores:
+            fio_job = fio.add_job(f"job_{core.core_id}")
```
Deixx: it's enough to do:
fio.add_job(...)
    .target(...)

PR #258, opened 2020-01-02
Add FI test: stop --no-flush-cache

Ostrokrzew:
Check that OpenCAS ‘stop -n’ option works correctly.
The test has been rewritten to meet current criteria.
Add error message to 'cli_messages' module.
Update tests using 'create_test_file' method due to rename that method.

Signed-off-by: Slawomir_Jankowski <slawomir.jankowski@intel.com>

### DIFF
--- a/test/functional/api/cas/cli_messages.py
+++ b/test/functional/api/cas/cli_messages.py
@@ -15,6 +15,14 @@ load_inactive_core_missing = [
     r"Successfully added cache instance \d+"
 ]
 
+start_cache_with_existing_metadata = [
+    r"Error inserting cache \d+",
+    r"Old metadata found on device\.",
+    r"Please load cache metadata using --load option or use --force to",
+    r" discard on-disk metadata and start fresh cache instance\.",
+    r"Error occurred, please see syslog \(/var/log/messages\) for details\."
+]
+
 remove_inactive_core = [
     r"Error while removing core device \d+ from cache instance \d+",
     r"Core device is in inactive state"
--- a/test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
+++ b/test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
@@ -1,0 +1,99 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+from time import sleep
+
+import pytest
+
+from api.cas import casadm, casadm_parser, cli, cli_messages
+from api.cas.cache_config import CacheMode, CleaningPolicy, CacheModeTrait
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from core.test_run import TestRun
+from test_tools import fs_utils
+from test_tools.dd import Dd
+from test_tools.disk_utils import Filesystem
+from test_utils import os_utils
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/cas"
+test_file_path = f"{mount_point}/test_file"
+
+
+@pytest.mark.parametrize("filesystem", Filesystem)
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stop_no_flush_load_cache(cache_mode, filesystem):
+    """
+        title: Test to check that 'stop --no-data-flush' command works correctly.
+        description: |
+          Negative test of the ability of CAS to load unflushed cache on core device
+          with filesystem. Test uses lazy flush cache modes.
+        pass_criteria:
+          - No system crash while load cache.
+          - Starting cache without loading metadata fails.
+          - Starting cache with loading metadata finishes with success.
+    """
+    with TestRun.step("Prepare cache and core devices."):
+        cache_part, core_part = prepare()
+
+    with TestRun.step("Start cache."):
+        cache = casadm.start_cache(cache_part, cache_mode, force=True)
+
+    with TestRun.step("Change cleaning policy to NOP."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
+
+    with TestRun.step(f"Add core with {filesystem.name} filesystem to cache and mount it."):
+        core_part.create_filesystem(filesystem)
+        core = cache.add_core(core_part)
+        core.mount(mount_point)
+
+    with TestRun.step(f"Create test file in mount point of exported object and check its md5 sum."):
+        test_file = fs_utils.create_random_test_file(test_file_path, Size(48, Unit.MebiByte))
+        test_file_md5_before = test_file.md5sum()
+
+    with TestRun.step("Unmount exported object."):
+        core.unmount()
+
+    with TestRun.step("Count dirty blocks on exported object."):
+        dirty_blocks_before = core.get_dirty_blocks()
+
+    with TestRun.step("Stop cache with option '--no-data-flush'."):
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try to start cache without loading metadata."):
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(
+            cache_dev=str(cache_part.system_path), cache_mode=str(cache_mode.name.lower()),
+            force=False, load=False))
+        cli_messages.check_msg(output, cli_messages.start_cache_with_existing_metadata)
+
+    with TestRun.step("Load cache."):
+        cache = casadm.load_cache(cache.cache_device)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 1:
+            TestRun.fail(f"Expected caches count: 1 Actual caches count: {caches_count}.")
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != 1:
+            TestRun.fail(f"Expected cores count: 1; Actual cores count: {cores_count}.")
+
+    with TestRun.step("Compare dirty blocks number before and after loading cache."):
+        if dirty_blocks_before != core.get_dirty_blocks():
+            TestRun.LOGGER.error("Dirty blocks number is different than before loading cache.")
+
+    with TestRun.step("Mount exported object."):
+        core.mount(mount_point)
+
+    with TestRun.step("Compare md5 sum of test file before and after loading cache."):
+        if test_file_md5_before != test_file.md5sum():
+            TestRun.LOGGER.error("Test file's md5 sum is different than before loading cache.")
+
+    with TestRun.step("Unmount exported object."):
+        core.unmount()
+
+    with TestRun.step("Stop cache."):
+        casadm.stop_all_caches()
+
--- a/test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
+++ b/test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
@@ -97,3 +97,80 @@ def test_stop_no_flush_load_cache(cache_mode, filesystem):
     with TestRun.step("Stop cache."):
         casadm.stop_all_caches()
 
+
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stop_no_flush_load_cache_no_fs(cache_mode):
+    """
+        title: Test to check that 'stop --no-data-flush' command works correctly.
+        description: |
+          Negative test of the ability of CAS to load unflushed cache on core device
+          without filesystem. Test uses lazy flush cache modes.
+        pass_criteria:
+          - No system crash while load cache.
+          - Starting cache without loading metadata fails.
+          - Starting cache with loading metadata finishes with success.
+    """
+    with TestRun.step("Prepare cache and core devices."):
+        cache_part, core_part = prepare()
+
+    with TestRun.step("Start cache with --force option."):
+        cache = casadm.start_cache(cache_part, cache_mode, force=True)
+
+    with TestRun.step("Change cleaning policy to NOP."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
+
+    with TestRun.step("Add core device without filesystem."):
+        core_part.wipe_filesystem()
+        core = cache.add_core(core_part)
+
+    with TestRun.step("Fill exported object with data."):
+        dd = (Dd()
+              .input("/dev/zero")
+              .output(core.system_path)
+              .block_size(Size(1, Unit.Blocks4096))
+              .oflag("direct"))
+        dd.run()
+
+    with TestRun.step("Count dirty blocks on exported object."):
+        dirty_blocks_before = core.get_dirty_blocks()
+
+    with TestRun.step("Stop cache with option '--no-data-flush'."):
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try to start cache without loading metadata."):
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(
+            cache_dev=str(cache_part.system_path), cache_mode=str(cache_mode.name.lower()),
+            force=False, load=False))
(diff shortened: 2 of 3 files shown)


test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,139 @@
...
+    with TestRun.step("Start proper tests"):
+        stop_n_load_cache_fs(cache_mode, filesystem, cache_dev, core_dev)
+        stop_n_load_cache_notfs(cache_mode, cache_dev, core_dev)
+
+
+def stop_n_load_cache_fs(cache_mode, filesystem, cache_dev, core_dev):
+    with TestRun.step("Start cache."):
+        cache = casadm.start_cache(cache_dev, cache_mode, force=True)
+
+    with TestRun.step("Add core to cache with xfs filesystem and mount it."):
```
micrakow: Isn't filesystem being parametrized rather than fixed to xfs?

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,139 @@
...
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try start cache without loading metadata."):
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(cache_dev=str(cache_dev),
+                                                                cache_mode=str(cache_mode),
+                                                                force=False, load=False))
+        if not output.stderr:
```
micrakow: How about checking if proper err msg is printed as well to be sure that starting cache failed because of expected reason?

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def stop_n_load_cache_notfs(cache_mode, cache_dev, core_dev):
+    with TestRun.step("Start cache."):
+        cache = casadm.start_cache(cache_dev, cache_mode, force=True)
+
+    with TestRun.step("Add core device without filesystem."):
+        cache.add_core(core_dev)
+
+    with TestRun.step("Copy data to cache device"):
```
micrakow: What for data is copied here?

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,139 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
```
micrakow: I did not intend to give my approve during last review :dancer:

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
Ostrokrzew: This is step from original test. I think it's to fill cache with data, which would not be flushed.
Also thanks, because I saw I done operation on wrong device.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,145 @@
...
+        core_dev.create_partitions([Size(2, Unit.GibiByte)])
+        core_dev = core_dev.partitions[0]
+
+    with TestRun.step("Start proper tests"):
+        stop_n_load_cache_fs(cache_mode, filesystem, cache_dev, core_dev)
+        stop_n_load_cache_notfs(cache_mode, cache_dev, core_dev)
+
+
+def stop_n_load_cache_fs(cache_mode, filesystem, cache_dev, core_dev):
+    with TestRun.step("Start cache."):
```
katlapinka: Those steps are nested in "with TestRun.step("Start proper tests"):" - it shouldn't be done like that. You can create helpers methods, but they should execute blocks of code without calling steps. Moreover, when all steps are placed in main test's body it is more readable and it would be a lot easier to parse it in the future, e.g. for the documentation purpose.

KamilLepek:
This patch contains updated submodule file explicitly in this repo (instead of submodule repo). This creates a problem with tests (unable to checkout submodule since on the submodule repo these changes do not exist).

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,135 @@
...
+    with TestRun.step("Unmount exported object."):
+        core.unmount()
+
+    with TestRun.step("Stop cache."):
+        cache.stop()
+
+    with TestRun.step("Wipe filesystem."):
+        core_dev.wipe_filesystem()
+
+    with TestRun.step("Start cache."):
```
Deixx: add 'with --force option'

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,135 @@
...
+              .oflag("direct"))
+        dd.run()
+
+    with TestRun.step("Stop cache with option '--no-data-flush'."):
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try start cache without loading metadata."):
```
Deixx: Try 'starting' or 'to start'

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,135 @@
...
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(cache_dev=str(cache_dev),
+                                                                cache_mode=str(cache_mode),
+                                                                force=False, load=False))
+        if output.stderr:
+            TestRun.LOGGER.info(output.stderr)
+        else:
+            TestRun.LOGGER.error(output.stdout)
+            TestRun.fail("Cache start succeeded (should fail)!")
+
+    with TestRun.step("Stop all caches."):
```
Deixx: it could be just 'Stop cache', despite method name

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,133 @@
...
+    with TestRun.step("Add core device without filesystem."):
+        core = cache.add_core(core_part)
+
+    with TestRun.step("Fill exported object with data."):
+        dd = (Dd()
+              .input("/dev/zero")
+              .output(f"{core.system_path}")
+              .block_size(Size(1, Unit.Blocks4096))
+              .oflag("direct"))
+        dd.run()
```
robertbaldyga: What is that `dd` for?

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,133 @@
...
+                                                                cache_mode=str(cache_mode),
+                                                                force=False, load=False))
+        if output.stderr:
+            TestRun.LOGGER.info(output.stderr)
+        else:
+            TestRun.LOGGER.error(output.stdout)
+            TestRun.fail("Cache start succeeded (should fail)!")
+
+    with TestRun.step("Stop cache."):
+        casadm.stop_all_caches()
```
robertbaldyga: This is not needed and it should never execute as even if cache had started in previous step, the test would have been immediately interrupted by `TestRun.fail()`.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,133 @@
...
+@pytest.mark.parametrize("cache_mode", CacheMode)
+@pytest.mark.parametrize("filesystem", Filesystem)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_stop_no_flush_load_cache(cache_mode, filesystem):
+    """
+        title: Test to check that 'stop --no-data-flush' command works correctly.
+        description: |
+          Negative test of the ability of the CAS to load unflushed cache
+          when core device is mounted and unmounted.
```
robertbaldyga: This test never tries to load cache when core device is mounted.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,133 @@
...
+        cache = casadm.start_cache(cache_part, cache_mode, force=True)
+
+    with TestRun.step(f"Add core with {filesystem.name} filesystem to cache and mount it."):
+        core_part.create_filesystem(filesystem)
+        core = cache.add_core(core_part)
+        core.mount(mount_point)
+
+    with TestRun.step(f"Create test file in /tmp directory and check its md5 sum."):
+        test_file = fs_utils.create_test_file(f'{test_file_path}',
+                                              'There is nothing interesting inside this file.')
```
robertbaldyga: It's so unsatisfying that we don't have in TF elegant method of generating reasonable* test files yet.
In this test this test file is practically meaningless and creating it could be removed without any loss, but in general this looks quite sad.

\* By reasonable I mean bigger than one sector containing joke string instead of verifiable pattern.

Ostrokrzew: I've changed this function in TF. I think its more reasonable now, but please, check it - https://github.com/Open-CAS/test-framework/pull/114.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,153 @@
...
+    with TestRun.step("Change cleaning policy to NOP."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
+
+    with TestRun.step(f"Add core with {filesystem.name} filesystem to cache and mount it."):
+        core_part.create_filesystem(filesystem)
+        core = cache.add_core(core_part)
+        core.mount(mount_point)
+
+    with TestRun.step(f"Create test file in mount point of exported object and check its md5 sum."):
+        test_file = fs_utils.create_random_test_file(f'{test_file_path}', Size(128, Unit.MebiByte))
```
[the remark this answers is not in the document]
Ostrokrzew: The purpose of this function in this place is to create some data on exported object. This would cause a dirty blocks in the cache. We would like to count this dirty blocks and check if its number is equal after stopping the cache without data flushing and load the cache again. If yes, than this option makes sense, if no, this test should fail.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+
+    with TestRun.step("Stop cache with option '--no-data-flush'."):
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try to start cache without loading metadata."):
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(
+            cache_dev=str(cache_part.system_path), cache_mode=str(cache_mode),
```
katlapinka: cache_mode.name, because correct value of the parameter is for example "wo", not "Write-Only" which is returned as this enum __str__() method

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+            force=False, load=False))
+        cli_messages.check_msg(output, cli_messages.start_wo_load_on_cache_dev_prev_used)
```
katlapinka: cli_messages.start_wo_load_on_cache_dev_prev_used - this is wrong expected message, because you passed wrong value of cache mode parameter. The message should be like (you should check it, because I paste here the example output):
"Error inserting cache 1
Old metadata found on device.
Please load cache metadata using --load option or use --force to
 discard on-disk metadata and start fresh cache instance.

Error occurred, please see syslog (/var/log/messages) for details. 
"

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+
+    with TestRun.step("Stop cache with option '--no-data-flush'."):
+        cache.stop(no_data_flush=True)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 0:
+            TestRun.fail(f"Expected caches count: 0; Actual caches count: {caches_count}.")
+
+    with TestRun.step("Try to start cache without loading metadata."):
+        output = TestRun.executor.run_expect_fail(cli.start_cmd(
+            cache_dev=str(cache_part.system_path), cache_mode=str(cache_mode),
```
katlapinka: same as above

Deixx:
retest this please

test/functional/api/cas/cli_messages.py
```diff
@@ -15,6 +15,14 @@
     r"Successfully added cache instance \d+"
 ]
 
+start_wo_load_on_cache_dev_prev_used = [
+    r"Error inserting cache \d+",
+    r"Old metadata found on device\.",
+    r"Please load cache metadata using --load option or use --force to",
+    r" discard on-disk metadata and start fresh cache instance\.",
```
Deixx: If this is one line in output, make it one line

test/functional/api/cas/cli_messages.py
```diff
@@ -15,6 +15,14 @@
     r"Successfully added cache instance \d+"
 ]
 
+start_wo_load_on_cache_dev_prev_used = [
```
Deixx: rename to sth more meaningful like start_cache_with_existing_metadata
BTW 'wo' can be misunderstood, especially that we have WO mode

Ostrokrzew: good point

test/functional/api/cas/cli_messages.py
Ostrokrzew: Actually it is not. This line is broken like that.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+from test_tools import fs_utils
+from test_tools.dd import Dd
+from test_tools.disk_utils import Filesystem
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/cas"
+test_file_path = f"{mount_point}/test_file"
+
+
+@pytest.mark.parametrize("cache_mode", [CacheMode.WB, CacheMode.WO])
```
katlapinka: Maybe instead of WB and WO use sth like:
@pytest.mark.parametrize("cache_mode", [mode for mode in CacheMode if
                                        CacheModeTrait.LazyFlush & CacheMode.get_traits(mode)])

in case there will be some new cache mode supporting lazy flush in the future.
Also please update the description few lines below.

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+            TestRun.fail(f"Expected caches count: 1 Actual caches count: {caches_count}.")
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != 1:
+            TestRun.fail(f"Expected cores count: 1; Actual cores count: {cores_count}.")
+
+    with TestRun.step("Mount exported object."):
+        core.mount(mount_point)
+
+    with TestRun.step("Compare dirty blocks number before and after loading cache."):
+        if dirty_blocks_before != cache.get_dirty_blocks():
```
katlapinka: here as well please count dirty blocks on core

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+
+    with TestRun.step("Fill exported object with data."):
+        dd = (Dd()
+              .input("/dev/zero")
+              .output(core.system_path)
+              .block_size(Size(1, Unit.Blocks4096)))
+        dd.run()
+
+    with TestRun.step("Count dirty blocks on exported object."):
+        dirty_blocks_before = cache.get_dirty_blocks()
```
katlapinka: and here

test/functional/tests/fault_injection/test_fault_injection_opencas_load.py
```diff
@@ -0,0 +1,152 @@
...
+        cache = casadm.load_cache(cache.cache_device)
+        caches_count = len(casadm_parser.get_caches())
+        if caches_count != 1:
+            TestRun.fail(f"Expected caches count: 1 Actual caches count: {caches_count}.")
+        cores_count = len(casadm_parser.get_cores(cache.cache_id))
+        if cores_count != 1:
+            TestRun.fail(f"Expected cores count: 1; Actual cores count: {cores_count}.")
+
+    with TestRun.step("Compare dirty blocks number before and after loading cache."):
+        if dirty_blocks_before != cache.get_dirty_blocks():
```
katlapinka: and here

PR #265, opened 2020-01-03
Add basic kedr test

micrakow:
Needs: https://github.com/Open-CAS/test-framework/pull/102

### DIFF
--- a/test/functional/tests/stress/test_kedr.py
+++ b/test/functional/tests/stress/test_kedr.py
@@ -1,0 +1,210 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+
+from datetime import timedelta
+
+from test_tools.kedr import Kedr, KedrProfile
+from api.cas import cas_module, installer, casadm
+from core.test_run import TestRun
+from test_utils import os_utils
+from test_utils.size import Size, Unit
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+
+
+@pytest.fixture(scope="module")
+def install_kedr():
+    TestRun.LOGGER.info("Checking if kedr is installed")
+    if not Kedr.is_installed():
+        TestRun.LOGGER.info("Installing kedr")
+        Kedr.install()
+
+
+@pytest.fixture(scope="function")
+def unload_modules():
+    TestRun.LOGGER.info("Check if CAS is installed")
+    if installer.check_if_installed():
+        TestRun.LOGGER.info("Unloading modules")
+        cas_module.unload_all_cas_modules()
+
+    TestRun.LOGGER.info("Stop kedr if it is running")
+    if Kedr.is_loaded():
+        Kedr.stop()
+
+    TestRun.LOGGER.info("Mounting debugfs")
+    os_utils.mount_debugfs()
+
+
+@pytest.mark.parametrize("module", cas_module.CasModule)
+def test_kedr_memleak_load_cas_module(module, unload_modules, install_kedr):
+    """
+    title: Loading modules with kedr started with 'memleak' configuration
+    description: Load and unload modules with kedr started to watch for memory leaks
+    pass_criteria:
+      - No memory leaks observed after loading and unloading module
+    """
+    with TestRun.step(f"Starting kedr against {module}"):
+        Kedr.start(module.value)
+
+    with TestRun.step(f"Loading {module}"):
+        os_utils.load_kernel_module(module.value)
+
+    with TestRun.step(f"Unloading {module}"):
+        os_utils.unload_kernel_module(module.value, os_utils.ModuleRemoveMethod.modprobe)
+
+    with TestRun.step(f"Checking for memory leaks for {module}"):
+        try:
+            Kedr.check_for_mem_leaks(module.value)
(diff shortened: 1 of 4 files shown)


test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,136 @@
...
+        cache_part = cache_device.partitions[0]
+
+    with TestRun.step("Preparing core device"):
+        core_device = TestRun.disks['core']
+        core_device.create_partitions([Size(1, Unit.GibiByte)])
+        core_part = core_device.partitions[0]
+
+    with TestRun.step("Unload CAS modules if needed"):
+        if os_utils.is_kernel_module_loaded(module.value):
+            cas_module.unload_all_cas_modules()
```
katlapinka: Isn't it done in fixture (unload_modules) before test?

micrakow: It actually is, but start_cache uses helpers for creating partitions, which I believe in turn calls 'isdct show' command which reloads modules - so after unloading during fixture modules are loaded once more at this point of test...
I could possibly move unloading modules from fixture to beginning of each test but I think it is ok the way it is right now.

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,136 @@
...
+        Kedr.start(module.value)
+
+    with TestRun.step(f"Loading {module}"):
+        os_utils.load_kernel_module(module.value)
+
+    with TestRun.step(f"Unloading {module}"):
+        os_utils.unload_kernel_module(module.value, os_utils.ModuleRemoveMethod.modprobe)
+
+    with TestRun.step(f"Checking for memory leaks for {module}"):
+        Kedr.check_for_mem_leaks(module.value)
```
katlapinka: please add try/except block and fail test in except

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,136 @@
...
+        cache.add_core(core_dev=core_part)
+
+    with TestRun.step("Stopping cache"):
+        cache.stop()
+
+    with TestRun.step(f"Unloading CAS modules"):
+        cas_module.unload_all_cas_modules()
+
+    with TestRun.step(f"Checking for memory leaks for {module.value}"):
+        Kedr.check_for_mem_leaks(module.value)
```
katlapinka: as above

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,209 @@
...
+from test_utils.size import Size, Unit
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+
+@pytest.fixture(scope="module")
+def install_kedr():
+    with TestRun.step("Checking if kedr is installed"):
+        if not Kedr.is_installed():
+            with TestRun.step("Installing kedr"):
```
katlapinka: This step is nested in other step. It shouldn't be done like this - You cas use simply TestRun.LOGGER.info() here.
Additionally please avoid putting steps in fixtures or other helpers methods. It is more readable when all steps are placed in test body. Please change it in the whole file.

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,210 @@
+#
+# Copyright(c) 2019 Intel Corporation
```
Deixx: 2020

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,210 @@
...
+
+    TestRun.LOGGER.info("Mounting debugfs")
+    os_utils.mount_debugfs()
+
+
+@pytest.mark.parametrize("module", cas_module.CasModule)
+def test_kedr_memleak_load_cas_module(module, unload_modules, install_kedr):
+    """
+    title: Loading modules with kedr started with 'memleak' configuration
+    description: Load and unload modules with kedr started to watch for memmory leaks
```
Deixx: memory

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,210 @@
...
+        Kedr.start(module.value, KedrProfile.FAULT_SIM)
+
+    with TestRun.step("Setting up fault simulation parameters"):
+        Kedr.setup_fault_injections()
+
+    with TestRun.step(f"Trying to load {module}"):
+        out = os_utils.load_kernel_module(module.value)
+        if out.exit_code == 0 \
+                or "Cannot allocate memory" not in out.stderr:
+            TestRun.fail(f"Loading module should fail because of alloc error, instead "
```
Deixx: Wouldn't it be better to not break the test here and allow stopping kedr in the next step? (in all tests)

test/functional/tests/stress/test_kedr.py
```diff
@@ -0,0 +1,210 @@
...
+
+@pytest.mark.parametrize("module", cas_module.CasModule)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_kedr_basic_io(module, unload_modules, install_kedr):
+    """
+    title: Basic IO test with kedr started with memory leaks profile
+    description: |
+        Load CAS modules, start kedr against one of them, start cache and add core,
+        run simple 4 minute random IO, stop cache and unload modules
```
Deixx: time different than in test

PR #268, opened 2020-01-08
Free memory of all allocated threads.

mmichal10:
This pach fixes memory leak which appeared due to commit d483951ebe6e040e53579.


modules/cas_cache/threads.c
```diff
@@ -144,7 +146,8 @@ static int _cas_metadata_updater_thread(void *data)
 				atomic_read(&info->kicked));
 	} while (true);
 
-	complete_and_exit(&info->compl, 0);
+	kfree(info);
```
robertbaldyga: This introduces potential race condition. Thread stopping is triggered by setting atomic variable `info->stop`, which is done in `_cas_stop_thread()`, and after that this function also tries to wake up `info->wq`. However if `_cas_stop_thread()` were called just before thread loop executed `wait_event_interruptible()`, the thread will never go to sleep and stop immediately freeing memory, hence `info` could be freed before `_cas_stop_thread()` tried wake up `info->wq`, effectively causing segmentation fault.

PR #271, opened 2020-01-13
tests: Add example test using power_control plugin

robertbaldyga:

### DIFF
--- a/test/functional/tests/example/example_test_power_cycle.py
+++ b/test/functional/tests/example/example_test_power_cycle.py
@@ -1,0 +1,26 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeSet
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_plugin("power_control")
+def test_create_example_partitions():
+    """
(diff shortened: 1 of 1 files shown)


test/functional/tests/example/example_test_power_cycle.py
```diff
@@ -0,0 +1,27 @@
...
+def test_create_example_partitions():
+    """
+        title: Example test doing power cycle
+        description: |
+          Example usage of power_control plugin.
+          NOTE:
+          This test uses plugin that is not included in test-framework.
+          It should be provided by user as external_plugin.
+        pass_criteria:
+          - Should pass.
```
katlapinka: DUT should reboot successfully?

test/functional/tests/example/example_test_power_cycle.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    with TestRun.step("Power cycle DUT"):
+        print(TestRun.plugins)
+        power_control = TestRun.plugin_manager.get_plugin('power_control')
+        power_control.power_cycle()
```
katlapinka: maybe add try-except block with TestRun.fail in except, so it wouldn't be exception as a test result, but fail

robertbaldyga: Exception thrown by `power_cycle()` method will also fail the test. Plus it will produce more meaningful message.

test/functional/tests/example/example_test_power_cycle.py
robertbaldyga: Fixed.

PR #273, opened 2020-01-14
Submodules update

Ostrokrzew:
Signed-off-by: Ostrokrzew <slawomir.jankowski@intel.com>


test/functional/.gitignore
```diff
@@ -1,3 +1,3 @@
-plugins/
+lib/external_plugins/
```
katlapinka: it is already changed and can be removed :)

PR #283, opened 2020-01-16
Modify error message

### LINKED ISSUE #280: Misleading error when running ./configure script without kernel-devel installed
When having kernel-headers package installed, but missing kernel-devel package Open CAS configure script shows misleading error:
```
# ./configure
Error: missing kernel headers
```
It would be expected for configure script to either shows error like 'missing kernel headers and/or kernel devel' or even better to distinguish between this two cases.

OS: RHEL 8.1
Kernel: 4.18.0-147.el8.x86_64

Ostrokrzew:
There is no generic way to check apart presence of kernel headers and kernel devel packages.
If we know system, we could use package manager, but without this knowledge,
we can't distinguish which headers in kernel headers directory belongs to which package.
Resolves #280 issue

### DIFF
--- a/configure
+++ b/configure
@@ -27,7 +27,7 @@ check_util lsblk
 
 if [ ! -e /lib/modules/$(uname -r)/build/ &> /dev/null ]
 then
-	echo >&2 "Error: missing kernel headers"
+	echo >&2 "Error: missing kernel headers and/or kernel devel"
 	MISSING_TOOLS=1
 fi
 


configure
```diff
@@ -25,12 +25,20 @@ check_util make
 check_util gcc
 check_util lsblk
 
-if [ ! -e /lib/modules/$(uname -r)/build/ &> /dev/null ]
+`rpm -qa | grep kernel-head &> /dev/null`
```
mmichal10: rpm is not installed by default on every os (i.e. ubuntu)

Ostrokrzew: Great thanks to You.

configure
```diff
@@ -27,7 +27,7 @@ check_util lsblk
 
 if [ ! -e /lib/modules/$(uname -r)/build/ &> /dev/null ]
 then
-	echo >&2 "Error: missing kernel headers"
+	echo >&2 "Error: missing kernel headers and/ or kernel devel"
```
Deixx: and/or

PR #284, opened 2020-01-16
Security test fix

Ostrokrzew:
Fix for creating partition on partition:
now partition would be created on parent device.

Without this change every execution of 'test_user_cli' from security tests would fail right after the start.

### DIFF
--- a/test/functional/tests/security/test_security_user.py
+++ b/test/functional/tests/security/test_security_user.py
@@ -5,9 +5,10 @@
 
 import pytest
 
-from cas import init_config, cli, casadm
-from cas.cache_config import CacheMode
-from cas.casadm_params import OutputFormat
+from api.cas import casadm, cli
+from api.cas.cache_config import CacheMode
+from api.cas.casadm_params import OutputFormat
+from api.cas.init_config import InitConfig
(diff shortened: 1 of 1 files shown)


test/functional/tests/security/test_security_user.py
```diff
@@ -41,8 +41,8 @@ def test_user_cli():
         cache_dev = cache_dev.partitions[0]
         core_dev = TestRun.disks['core']
         core_dev.create_partitions([Size(1, Unit.GibiByte), Size(256, Unit.MebiByte)])
-        core_dev = core_dev.partitions[0]
         core_dev1 = core_dev.partitions[1]
+        core_dev = core_dev.partitions[0]
```
katlapinka: What do you think about naming these variables like core_disk/cache_disk and those with assigned partitions like core_dev/cache_dev?

robertbaldyga: Why don't just name them cache_dev/core_dev and cache_part/core_part?

Ostrokrzew: That's good idea. Done.

Ostrokrzew:
retest this please

PR #288, opened 2020-01-20
Fix Python multiplication issue

rafalste:
When some particular values are multplied by 10^n, Python returns float
number with subtracted value of (size of float)^-1
For example: 1.017 * 1000 = 1016.(9)
Rounding the output value fixes this issue.

### DIFF
--- a/test/functional/api/cas/cache.py
+++ b/test/functional/api/cas/cache.py
@@ -177,7 +177,7 @@ class Cache:
                                               if alru_params.staleness_time else None,
                                               alru_params.flush_max_buffers
                                               if alru_params.flush_max_buffers else None,
-                                              int(alru_params.activity_threshold.total_seconds()
+                                              round(alru_params.activity_threshold.total_seconds()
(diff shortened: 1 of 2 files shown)


test/functional/api/cas/cache.py
```diff
@@ -177,7 +177,7 @@ def set_params_alru(self, alru_params: FlushParametersAlru):
                                               if alru_params.staleness_time else None,
                                               alru_params.flush_max_buffers
                                               if alru_params.flush_max_buffers else None,
-                                              int(alru_params.activity_threshold.total_seconds()
+                                              round(alru_params.activity_threshold.total_seconds()
```
robertbaldyga: I think the core problem here is that `total_seconds()` returns floating point number, so casting result of this function to int would allow to narrow down problem to it's actual root case, plus it prevents potential error accumulation (resulting from the nature of multiplication).

rafalste:
Replaced by PR #294

PR #310, opened 2020-01-28
Fix flush interrupting

mmichal10:


modules/cas_cache/layer_cache_management.c
```diff
@@ -204,11 +238,16 @@ static void _cache_mngt_cache_flush_complete(ocf_cache_t cache, void *priv,
 
 	result = _cache_mngt_async_callee_set_result(context, error);
 
-	if (result == -KCAS_ERR_WAITING_INTERRUPTED)
+	if (context->compl_func)
+		context->compl_func(cache);
```
arutk: At this point context might be freed. You must either 
1. call compl_func before _cache_mngt_async_callee_set_result or 
2. call compl_func  conditionally if result == -KCAS_ERR_WAITING_INTERRUPTED and perform equivalent operations in caller for the case of  result != -KCAS_ERR_WAITING_INTERRUPTED

modules/cas_cache/layer_cache_management.c
```diff
@@ -1071,8 +1160,11 @@ int _cache_mngt_remove_core_prepare(ocf_cache_t cache, ocf_core_t core,
 	if (!cmd->force_no_flush) {
 		if (core_active) {
 			/* Flush core */
-			flush_result = _cache_mngt_core_flush_sync(core,
-					flush_interruptible);
+			if (flush_interruptible)
+				flush_result = _cache_mngt_core_flush_sync(core,
```
arutk: What if the call gets interrupted and we exit this function and unlock & put the cache (in caller function) before async flush completes? Isn't this the exact scenario that you are attempting to fix ?

mmichal10: I added second cache_get() and left read_unlock() only in flush completion

PR #312, opened 2020-01-29
Fix for io class file classification tests

katlapinka:


test/functional/tests/io_class/test_io_class_file.py
```diff
@@ -448,4 +450,6 @@ def restore_classification_config():
     remove_files_classification()
 
     restore_classification_config()
+    core.unmount()
```
mmichal10: Please add a comment, why there should be umount and mount

test/functional/tests/io_class/test_io_class_file.py
```diff
@@ -420,8 +423,7 @@ def restore_classification_config():
         load_file_size_io_classes()
 
     cache, core = prepare()
-    Udev.disable()
-    base_size = Size(random.randint(50, 1000) * 2, Unit.Blocks4096)
+    base_size = Size(404, Unit.Blocks4096)
```
mmichal10: I would leave random size

test/functional/tests/io_class/test_io_class_file.py
```diff
@@ -137,7 +137,7 @@ def test_ioclass_file_name_prefix():
         dd.run()
         sync()
         current_occupancy = cache.get_occupancy()
-        assert current_occupancy == previous_occupancy
+        assert current_occupancy == previous_occupancy + (dd_size * dd_count)
```
mmichal10: Shouldn't you use TestRun.error or TestRun.fail instead of assert?

test/functional/tests/io_class/test_io_class_file.py
```diff
@@ -155,7 +155,8 @@ def test_ioclass_file_name_prefix():
         dd.run()
         sync()
         current_occupancy = cache.get_occupancy()
-        assert current_occupancy != previous_occupancy
+        assert current_occupancy == previous_occupancy
```
mmichal10: Shouldn't you use TestRun.error or TestRun.fail instead of assert?

PR #316, opened 2020-01-30
Functional test for acp cleaning policy

katlapinka:


test/functional/tests/write_back/test_wb_acp.py
```diff
@@ -0,0 +1,171 @@
...
+                              sector_in_chunk(c[0], header.sector_number)), None)
+                if not chunk:
+                    TestRun.LOGGER.error(f"Sector {header.sector_number} "
+                                         f"({Size(header.sector_number, Unit.Blocks512)}) "
+                                         f"not in current bucket.")
+                    all_writes_ok = False
+                    if all(lambda s: not sector_in_chunk(s[0], header.sector_number, chunk_list)):
+                        TestRun.LOGGER.warning(f"Sector {header.sector_number} "
+                                               f"({Size(header.sector_number, Unit.Blocks512)}) "
+                                               f"outside of any tested chunk.")
```
Deixx: Hmm, you're first logging an error for chunk not in current bucket and then just warning if it's not in all chunks, and I wonder if it's the right way to go. Do we want just warning for sectors written outside of all chunks (if that happens then it's not an ACP write) or do we treat it as a proper error (do we expect and allow any writes besides those generated by ACP?)

test/functional/tests/write_back/test_wb_acp.py
```diff
@@ -0,0 +1,171 @@
...
+from storage_devices.disk import DiskTypeSet, DiskTypeLowerThan, DiskType
+from test_tools.blktrace import BlkTrace, BlkTraceMask, ActionKind, RwbsKind
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import IoEngine, ReadWrite
+from test_utils.size import Size, Unit
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_wb_acp_functional():
```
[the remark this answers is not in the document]
Deixx: both test names and file/directory can be renamed to reflect that

test/functional/tests/write_back/test_wb_acp.py
```diff
@@ -0,0 +1,165 @@
...
+
+                # Get chunk within current bucket where current header sector is expected
+                chunk = next((c for c in bucket_chunks if
+                              sector_in_chunk(c[0], header.sector_number)), None)
+                if not chunk:
+                    TestRun.LOGGER.error(f"Sector {header.sector_number} "
+                                         f"({Size(header.sector_number, Unit.Blocks512)}) "
+                                         f"not in current bucket.")
+                    all_writes_ok = False
+                    if all([not sector_in_chunk(s[0], header.sector_number) for s in chunk_list]):
```
Deixx: For me more clear would be:
`if not any([sector_...`

test/functional/tests/lazy_flush/cleaning_policy/test_acp.py
```diff
@@ -0,0 +1,165 @@
...
+
+                # Get chunk within current bucket where current header sector is expected
+                chunk = next((c for c in bucket_chunks if
+                              sector_in_chunk(c[0], header.sector_number)), None)
+                if not chunk:
+                    TestRun.LOGGER.error(f"Sector {header.sector_number} "
+                                         f"({Size(header.sector_number, Unit.Blocks512)}) "
+                                         f"not in current bucket.")
+                    all_writes_ok = False
+                    if not any([sector_in_chunk(s[0], header.sector_number) for s in chunk_list]):
```
mmichal10: How about moving it to separate function `sector_in_tested_region()`?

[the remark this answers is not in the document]
katlapinka: renamed to c because chunk is already used in this scope

PR #321, opened 2020-02-04
Lazy-writes modes tests

Ostrokrzew:
Add tests for modes with lazy-writes: stop cache, remove core with filesystem from cache, remove core without filesystem from cache.

***WbCleanShutdown*** - clean cache stop:
```
StartCache,
AddCore,
WriteToCasDev,
ReadFromCasDev,
TestIfTheSame,
ReadFromCoreDev,
TestIfDifferent,
StopCache,
ReadFromCoreDev,
TestIfTheSame
```

***WbFsCleanShutdownRmCore*** - clean remove core with filesystem from cache:
```
StartCache,
AddCore,
WriteToCasDev,
ReadFromCasDev,
TestIfTheSame,
ReadFromCoreDev,
TestIfDifferent,
StopCache,
ReadFromCoreDev,
TestIfTheSame
```

***WbCleanShutdownRmCore*** - clean remove core without filesystem from cache:
```
StartCache,
AddCore,
WriteToCasDev,
ReadFromCasDev,
TestIfTheSame,
ReadFromCoreDev,
TestIfDifferent,
RemoveCore,
ReadFromCoreDev,
TestIfTheSame
```

Additionally I check if writes occurs there, where they should occur.

### DIFF
--- a/test/functional/tests/lazy_writes/test_lazy_writes_clean.py
+++ b/test/functional/tests/lazy_writes/test_lazy_writes_clean.py
@@ -152,6 +152,91 @@ def test_clean_stop_cache(cache_mode):
         test_file_3.remove(True)
 
 
+@pytest.mark.parametrize("fs", Filesystem)
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_clean_remove_core_with_fs(cache_mode, fs):
+    """
+        title: Test of the ability to remove core from cache in lazy-write modes with filesystem.
+        description: |
+          Test if OpenCAS removes core from cache in modes with lazy writes and with different
+          filesystems without data loss.
+        pass_criteria:
+          - Core removing works properly.
+          - Data on core device is correct after core is removed.
+    """
+    with TestRun.step("Prepare devices for cache and core."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(256, Unit.MebiByte)])
+        cache_part = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_dev.create_partitions([Size(512, Unit.MebiByte)])
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step(f"Add core with {fs.name} filesystem to cache and mount it."):
+        core_part.create_filesystem(fs)
+        core = cache.add_core(core_part)
+        core.mount(mnt_point)
+
+    with TestRun.step("Disable cleaning and sequential cutoff."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
+        cache.set_seq_cutoff_policy(SeqCutOffPolicy.never)
+
+    with TestRun.step("Create test file and read its md5 sum."):
+        test_file_main = create_random_test_file("/tmp/test_file_main", Size(64, Unit.MebiByte))
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Copy test file to the exported object."):
+        test_file_1 = File.create_file(mnt_point + "test_file_1")
+        dd = Dd().output(test_file_1.full_path) \
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_1.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_1.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Unmount and remove core."):
+        core.unmount()
+        core.remove_core()
+
+    with TestRun.step("Mount core device."):
+        core_part.mount(mnt_point)
+
+    with TestRun.step("Read data from the core device."):
+        test_file_2 = File.create_file("/tmp/test_file_2")
+        dd = Dd().output(test_file_2.full_path) \
+            .input(test_file_1.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_1.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_2.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Delete test files."):
+        test_file_main.remove(True)
+        test_file_1.remove(True)
+        test_file_2.remove(True)
+
+    with TestRun.step("Unmount core device."):
+        core_part.unmount()
+        remove(mnt_point, True, True, True)
+
 
 
 def check_device_write_stats(device: Device):
--- a/test/functional/tests/lazy_writes/test_lazy_writes_clean.py
+++ b/test/functional/tests/lazy_writes/test_lazy_writes_clean.py
@@ -238,6 +238,137 @@ def test_clean_remove_core_with_fs(cache_mode, fs):
         remove(mnt_point, True, True, True)
 
 
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_clean_remove_core_without_fs(cache_mode):
+    """
+        title: Test of the ability to remove core from cache in lazy-write modes without filesystem.
+        description: |
+          Test if OpenCAS removes core without filesystem in modes with lazy writes
+          without data loss.
+        pass_criteria:
+          - Core removing works properly.
+          - Writes to exported object and core device during OpenCAS's work are equal
+          - Data on core device is correct after core is removed.
+    """
+    with TestRun.step("Prepare devices for cache and core."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([Size(256, Unit.MebiByte)])
+        cache_part = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        core_dev.create_partitions([Size(512, Unit.MebiByte)])
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+
+    with TestRun.step("Disable cleaning and sequential cutoff."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
+        cache.set_seq_cutoff_policy(SeqCutOffPolicy.never)
+
+    with TestRun.step("Read IO stats before test"):
+        core_disk_writes_initial = check_device_write_stats(core_part)
+        exp_obj_writes_initial = check_device_write_stats(core)
+
+    with TestRun.step("Write data to exported object."):
+        test_file_main = create_random_test_file("/tmp/test_file_main", Size(64, Unit.MebiByte))
+        dd = Dd().output(core.system_path) \
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Read IO stats after write to the exported object."):
+        core_disk_writes_increase = (
+            check_device_write_stats(core_part) - core_disk_writes_initial
+        )
+        exp_obj_writes_increase = (
+            check_device_write_stats(core) - exp_obj_writes_initial
+        )
+
+    with TestRun.step("Validate IO stats after write to the exported object."):
+        if core_disk_writes_increase > 0:
+            TestRun.LOGGER.error("Writes should occur only on the exported object.")
+        if exp_obj_writes_increase != test_file_main.size.value:
+            TestRun.LOGGER.error("Not all writes reached the exported object.")
+
+    with TestRun.step("Read data from the exported object."):
+        test_file_1 = File.create_file("/tmp/test_file_1")
+        dd = Dd().output(test_file_1.full_path) \
+            .input(core.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_1.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_1.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Read data from the core device."):
+        test_file_2 = File.create_file("/tmp/test_file_2")
+        dd = Dd().output(test_file_2.full_path) \
+            .input(core_part.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_2.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main == test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be different.")
+
+    with TestRun.step("Read IO stats before removing core."):
+        core_disk_writes_before_remove = check_device_write_stats(core_part)
+
+    with TestRun.step("Remove core."):
+        core.remove_core()
+
+    with TestRun.step("Read IO stats after removing core."):
+        core_disk_writes_increase = (
+            check_device_write_stats(core_part) - core_disk_writes_before_remove
+        )
+
+    with TestRun.step("Validate IO stats after removing core."):
+        if core_disk_writes_increase == 0:
+            TestRun.LOGGER.error("Writes should occur on the core device after removing core.")
+        if core_disk_writes_increase != exp_obj_writes_increase:
+            TestRun.LOGGER.error("Write statistics for the core device should be equal "
+                                 "to those from the exported object.")
+
+    with TestRun.step("Read data from core device again."):
+        test_file_3 = File.create_file("/tmp/test_file_3")
+        dd = Dd().output(test_file_3.full_path) \
+            .input(core_part.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_3.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_3.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Delete test files."):
+        test_file_main.remove(True)
+        test_file_1.remove(True)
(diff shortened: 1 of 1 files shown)


test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+        dd = Dd().output(core.system_path) \
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+
+    with TestRun.step("Check IO stats."):
+        # 4kB is written when partition is created
+        core_stats = (IOstat.get_iostat(core_part).total_writes - Size(4, Unit.KibiByte)).value
```
katlapinka, line 63: You should disable seq cut-off to make sure that statistics have proper values, because dd is runnning sequential I/O, which after gaining  set threshold will be run in pass-through

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_1.refresh_item()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_main.md5sum() != test_file_1.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Read data from core device."):
```
katlapinka, line 82: I recommend to stop cache before reading data from core device, because there are cases, when system is making md5 differ when cache is running, but after stop everything is ok. What is more, test is for cache modes with lazy writes cache mode, so you should flush data before checking md5sums on core and core device.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+
+    with TestRun.step("Change cleaning policy to 'NOP'."):
+        cache.set_cleaning_policy(CleaningPolicy.nop)
```
katlapinka, line 154: disable seq cut-off

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_1.refresh_item()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_main.md5sum() != test_file_1.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Read data from core device."):
```
katlapinka, line 186: remove core before checking md5sums

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+            TestRun.LOGGER.error("Md5 sums should be different.")
+
+    with TestRun.step("Remove core."):
+        core.remove_core()
+
+    with TestRun.step("Check IO stats again."):
+        # 4kB is written when partition is created
+        core_stats = (IOstat.get_iostat(core_part).total_writes - Size(4, Unit.KibiByte)).value
+        if core_stats == 0 or core_stats != exp_obj_stats:
+            TestRun.LOGGER.error("Writes should occur only on cache device.")
```
mmichal10, line 207: rename `core_stats` and split `if` into to separate with appropriate error message

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+        if test_file_main.md5sum() == test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be different.")
+
+    with TestRun.step("Remove core."):
+        core.remove_core()
+
+    with TestRun.step("Check IO stats again."):
+        # 4kB is written when partition is created
+        core_stats = (IOstat.get_iostat(core_part).total_writes - Size(4, Unit.KibiByte)).value
+        if core_stats == 0 or core_stats != exp_obj_stats:
```
mmichal10, line 206: You just removed core with flush, so writes to core disk are expected

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
mmichal10: I think core is removed in good order. At first there is a check if data written to attached core is correct. Integrity after core removing is checked few steps later.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+            .input(core_part.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_3.refresh_item()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_main.md5sum() != test_file_3.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
```
mmichal10, line 221: It you removed core with flushing dirty data, shouldn't md5 be the same?

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,230 @@
...
+        test_file_main.remove(True)
+        test_file_1.remove(True)
+        test_file_2.remove(True)
+        test_file_3.remove(True)
+
+
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_clean_remove_core(cache_mode):
```
katlapinka, line 129: There should be a filesystem on core device and core device should be mounted!

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,259 @@
...
+        test_file_main.refresh_item()
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Check IO stats."):
+        core_disk_writes = IOstat.get_iostat(core_part).total_writes.value - core_disk_writes_diff
+        exp_obj_stats = IOstat.get_iostat(core).total_writes.value - exp_obj_stats_diff
+        if core_disk_writes > 0:
+            TestRun.LOGGER.error("Writes should occur only on cache device.")
+        if exp_obj_stats != test_file_main.size.value:
+            TestRun.LOGGER.error("Not all writes reached exported object.")
```
mmichal10, line 192: You have a FS on core device. Please run this test for each FS and check if there is no inconsistency in stats caused by filesystem metadata writes. You can workaround it by adding ioclass for metadata with disabled selective allocation.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,259 @@
...
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_clean_remove_core(cache_mode):
+    """
+        title: Test of the ability to remove core from cache in modes with lazy writes.
+        description: |
+          Test if OpenCAS removes core in modes with lazy writes without data loss.
+        pass_criteria:
+          - Core removing works properly.
+          - Data on core device is correct after core is removed.
```
mmichal10, line 148: Mention about filesystem in testname/description/titile

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,259 @@
...
+            .input(core_part.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_2.refresh_item()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main == test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be different.")
```
mmichal10, line 220: You are reading raw data from device with filesystem, so this check doesn't have to much sense.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -231,86 +231,119 @@ def test_clean_remove_core_with_fs(cache_mode, fs):
...
 
 
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_clean_remove_core_without_fs(cache_mode):
+    """
+        title: Test of the ability to remove core from cache in modes with lazy writes.
+        description: |
+          Test if OpenCAS removes core with in modes with lazy writes without data loss.
```
Deixx, line 241: add info to the title and description, that the test handles the case without fs.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -231,86 +231,119 @@ def test_clean_remove_core_with_fs(cache_mode, fs):
...
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+        # get IO stats independent from test
         core_disk_writes_diff = IOstat.get_iostat(core_part).total_writes.value
```
Deixx, line 261: change name - this is not a diff (unless you treat 0 as the value for differentiation)

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,343 @@
...
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Check IO stats."):
+        # get IO stats generated by test
+        core_disk_writes = (IOstatBasic.get_iostat_list([core_part])[0].total_writes.value
```
Deixx, line 70: So, core_disk_writes is the DIFFERENCE between the current value of writes and the initial value of writes after creating partitions (which you named _diff). Current names do not reflect the actual values of those variables.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,343 @@
...
+        test_file_2.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Check IO stats again."):
+        # get IO stats generated by test
+        core_disk_writes = (IOstatBasic.get_iostat_list([core_part])[0].total_writes.value
```
Deixx, line 114: same here

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,343 @@
...
+        core_dev.create_partitions([Size(512, Unit.MebiByte)])
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+        # get IO stats independent from test - adding core creates metadata on core device
```
Deixx, line 249: adding core does not create metadata on core

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,343 @@
...
+        core_disk_writes_diff = IOstatBasic.get_iostat_list([core_part])[0].total_writes.value
```
Deixx, line 250: As I wrote before - diff means the difference between some values. I know I wrote "unless you treat 0 as the value for differentiation", and that left an option for leaving this name as is, but I'd prefer renamed variable.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,343 @@
...
+        core_dev.create_partitions([Size(512, Unit.MebiByte)])
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+        # get IO stats independent from test - adding core creates metadata on core device
```
Deixx, line 51: adding core does not create metadata on core

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+        test_file_main = create_random_test_file("/tmp/test_file_main", Size(64, Unit.MebiByte))
+        dd = Dd().output(core.system_path) \
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Check IO stats generated by previous step."):
```
Deixx, line 71: Generally - don't use expressions like "previous step" in step description. If test ever needs to be modified and a step is added before this, then it would be easy to miss that. "Check IO stats increase." would be sufficient

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+            .oflag("direct")
+        dd.run()
+        test_file_2.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if not is_equal(test_file_md5sum_main, test_file_2.md5sum()):
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Check IO stats generated for the core device since last check."):
```
Deixx, line 117: This is misleading - I don't see anywhere reassigning core_disk_writes_initial, so you are checking increase in relation to initial value, and not since the last check.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step(f"Add core with {fs.name} filesystem to cache and mount it."):
+        core_part.create_filesystem(fs)
+        core = cache.add_core(core_part)
+        core.mount(mnt_point)
+
+    with TestRun.step("Disable cleaning and sequential cutoff policies."):
```
Deixx, line 165: You cannot disable policies - you can change policies, thus disabling cleaning and sequential cutoff. So the word "policies" should be removed.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+        core_part = core_dev.partitions[0]
+        Udev.disable()
+
+    with TestRun.step(f"Start cache in {cache_mode} mode."):
+        cache = casadm.start_cache(cache_part, cache_mode)
+
+    with TestRun.step("Add core to cache."):
+        core = cache.add_core(core_part)
+
+    with TestRun.step("Disable cleaning and sequential cutoff policies."):
```
Deixx, line 255: remove "policies"

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+        test_file_main = create_random_test_file("/tmp/test_file_main", Size(64, Unit.MebiByte))
+        dd = Dd().output(core.system_path) \
+            .input(test_file_main.full_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_md5sum_main = test_file_main.md5sum()
+
+    with TestRun.step("Check IO stats generated by previous step."):
```
Deixx, line 273: same as above

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+        core_disk_writes_current = (
```
Deixx, line 274: same as above

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,369 @@
...
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if is_equal(test_file_md5sum_main, test_file_2.md5sum()):
+            TestRun.LOGGER.error("Md5 sums should be different.")
+
+    with TestRun.step("Remove core."):
+        core.remove_core()
+
+    with TestRun.step("Check IO stats generated for th core device since last check."):
```
Deixx, line 319: same as above (step title)

Ostrokrzew:
retest this please

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
Deixx: Sorry - before I focused on step description matching actual action and not on which is correct. It looks like you want to check here if there were writes during cache stop. In that case, the previous description was ok, but the code was not doing that.

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
Deixx: I made the same mistake here

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,382 @@
...
+            .oflag("direct")
+        dd.run()
+        test_file_2.refresh_item()
+        sync()
+
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+        if test_file_1.md5sum() != test_file_2.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
```
Deixx, line 234: unnecessary - you're checking earlier if test_file_md5sum_main equals test_file_1.md5sum()

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,382 @@
...
+    with TestRun.step("Compare md5 sum of test files."):
+        if test_file_md5sum_main != test_file_1.md5sum():
+            TestRun.LOGGER.error("Md5 sums should be equal.")
+
+    with TestRun.step("Unmount and remove core."):
+        core.unmount()
+        core.remove_core()
+
+    with TestRun.step("Stop cache."):
+        cache.stop()
```
Deixx, line 214: unnecessary

test/functional/tests/lazy_writes/test_lazy_writes_clean.py
```diff
@@ -0,0 +1,382 @@
...
+            .input(core_part.system_path) \
+            .block_size(bs) \
+            .count(int(test_file_main.size / bs)) \
+            .oflag("direct")
+        dd.run()
+        test_file_3.refresh_item()
+        sync()
+
+    with TestRun.step("Stop cache."):
+        cache.stop()
```
Deixx, line 367: unnecessary

Ostrokrzew:
retest this please

PR #331, opened 2020-02-11
Load cache tests' update

Ostrokrzew:
Refactor load tests.
Extend cleaning policy params' classes.


test/functional/api/cas/cache_config.py
```diff
@@ -147,6 +147,12 @@ def alru_params_range():
         alru_params.wake_up_time = (1, 3600)
         return alru_params
 
+    def __eq__(self, other):
```
mmichal10: I believe 'Overload equality operator' sounds better than 'Add equation overload'

PR #344, opened 2020-03-03
Add package generator script

rafalste:
- add a 'pckgen' script to generate various OpenCAS packages like
release archives (tar, zip), RPMs (source and binary) as well as to
create version file with metadata
- add a SPEC file for RPM creation
- add Makefile targets to use 'pckgen' script through 'make \<package\>'

'pckgen' script has a bunch of features for package generation.
Read './pckgen --help' for more information on usage scenarios.

Script is also prepared to easily add new packages creation in the future.

### DIFF
--- a/utils/pckgen
+++ b/utils/pckgen
@@ -1,0 +1,468 @@
+#!/bin/bash
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+# To make additional packages creation possible
+# through this script all you need to do is:
+# - add 'case' command to set variable 'GENERATE_<NAME>=1'
+# - add function 'generate_<name>' to handle the task
+# - add 'DEPENDENCIES_<NAME>' array if needed
+# - update help!
+
+
+NAME="open-cas-linux"
+SUPPORTED_FROM_VERSION="20.3"
+THIS="$(basename $0)"
+ARCH="$(uname -i)"
+SCRIPT_BASE_DIR="$(dirname $(realpath $0))"
+BASE_SPEC="$SCRIPT_BASE_DIR/$NAME.spec.base"
+TEMP_TEMPLATE="opencas-${THIS}"
+DEPENDENCIES=(sed git)
+# Dependencies for particular packages creation:
+DEPENDENCIES_TAR=(tar)
+DEPENDENCIES_ZIP=(zip)
+DEPENDENCIES_RPM=(tar rpmbuild)
+DEPENDENCIES_SRPM=("${DEPENDENCIES_RPM[@]}")
+# List of relative submodule directories:
+SUBMODULES=(
+    "ocf"
+    "test/functional/test-framework"
+)
+
+
+usage() {
+    echo "Usage:"
+    echo "  ./$THIS <commands> [options] <SOURCES_PATH>"
+    echo "  ./$THIS -c|-h"
+    echo "where SOURCES_PATH is the root directory of OpenCAS sources"
+}
+
+print_help() {
+    echo "Generate OpenCAS packages."
+    echo "$(usage)"
+    echo
+    echo "This script generates various OpenCAS packages like release archives (tar, zip),"
+    echo "RPMs (source and binary) as well as creates version file with metadata."
+    echo
+    echo "Mandatory arguments to long options are mandatory for short options too."
+    echo
+    echo "Commands:"
+    echo "  tar                             generate tar archive"
+    echo "  zip                             generate zip archive"
+    echo "  rpm                             generate RPM packages"
+    echo "  srpm                            generate SRPM package"
+    echo "  version-only                    create only version file and exit;"
+    echo
+    echo "Options:"
+    echo "  -o, --output-dir <DIR>          put all created files in the given directory;"
+    echo "                                  by default all files (except version file)"
+    echo "                                  will be put in 'SOURCES_PATH/packages/'"
+    echo "  -d, --debug                     create all debug files and packages as well"
+    echo "  -c, --clean                     clean all temporary files and folders that"
+    echo "                                  may have been left around if $THIS ended"
+    echo "                                  unexpectedly in the previous run"
+    echo "  -h, --help                      print this help message"
+    echo
+}
+
+invalid_usage() {
+    >&2 echo -e "$THIS: $*\nTry './$THIS --help' for more information."
+    exit 2
+}
+
+info() {
+    echo -e "\e[33m$*\e[0m"
+}
+
+success() {
+    echo -e "\n\e[32m=== ALL DONE ===\e[0m\n\nYou can find your fresh packages in '$OUTPUT_DIR'\n"
+}
+
+error() {
+    >&2 echo -e "\e[31mERROR\e[0m: $THIS: $*"
+    exit 1
+}
+
+clean() {
+    rm -rf "$TEMP_DIR"
+    if [ -d "$TEMP_DIR" ]; then
+        # Print only warning to not confuse the user by an error here
+        # if packages were created successfully and everything else
+        # went fine except cleaning temp directory at the end.
+        info "WARNING: Cleanup failed"
+    fi
+}
+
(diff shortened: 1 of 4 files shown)


utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+SUBMODULES=(
+    "ocf"
+    "test/functional/test-framework"
+)
+
+
+usage() {
+    echo "Usage:"
+    echo "  ./$THIS -t -z -r -s [options] <SOURCES_PATH>"
+    echo "  ./$THIS -v [options] <SOURCES_PATH>"
```
arutk: is <SOURCE_PATH> required when a ready input tarball is provided (-f or -u) ?

utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+        cp -fat "$TEMP_SOURCES_DIR" "$SOURCES_DIR/."
+
+        if [ -d "$TEMP_SOURCES_DIR/.git" ]; then
+            echo "--- Updating submodules"
+            for SUBMOD in ${SUBMODULES[@]}; do
+                if [ ! -f "$TEMP_SOURCES_DIR/$SUBMOD/.git" ]; then
+                    local SUBMODULES_UPDATE=1
+                fi
+            done
+            if [ ! $SUBMODULES_UPDATE ] && [ ! $FORCE ]; then
```
arutk: I'm not sure whether I get this right.

If you run the script without --force then :
a) if all modules are initialized, you are going to print a warning "Submodules already initialized (use --force to update)"
b) if some modules are not initialized, you don't print anything, intialize the ones that are missing and update all of them silently.

This seems inconsistent

arutk: Actually why would you ever want to update modules? I would just make it caller responsibility to have the modules in the right version.  If you skip updating modules, you can copy the directory without ".git" subdirectories, so you don't need to remove them in the following steps.

utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+        tar -C "$TEMP_DIR" -zcf "$OUTPUT_DIR/$SOURCES_TAR_NAME" "$CAS_FILENAME"
+        if [ $? -ne 0 ] || [ ! -f "$OUTPUT_DIR/$SOURCES_TAR_NAME" ]; then
+            rm -rf "$OUTPUT_DIR/$SOURCES_TAR_NAME"
+            error "Couldn't create tar archive"
+        fi
+    fi
+}
+
+generate_zip() {
+    echo "--- Creating zip archive from current sources"
```
arutk: In generate_tar() you check whether the output file already exists. Here you don't. Please make this behavior consistent across all functions.

utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+
+        ARCHIVE_PREPARED=1
+    fi
+}
+
+generate_tar() {
+    echo "--- Creating tar archive from current sources"
+    if [ -f "$OUTPUT_DIR/$SOURCES_TAR_NAME" ] && [ ! $FORCE ]; then
+        echo "---   Tarball already present in output folder (use --force to overwrite)"
+    else
```
robertbaldyga: Typically we don't care. When we run `make`, the new binaries will just replace the old ones. Why should we care in case of tar package?

utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+    echo "  -t, --make-tar                  generate tar archive"
+    echo "  -z, --make-zip                  generate zip archive"
+    echo "  -r, --make-rpm                  generate RPM packages"
+    echo "  -s, --make-srpm                 generate SRPM package"
+    echo "  -v, --version-only              create only version file and exit;"
+    echo "                                  by default in 'SOURCES_PATH/metadata' directory,"
+    echo "                                  use -o option to redirect"
+    echo "  -M, --master                    if building packages, make them 'master' version;"
+    echo "                                  $THIS creates 'devel' version by default"
+    echo "  -R, --release                   if building packages, make them 'release' version;"
```
robertbaldyga: We don't need this flag at all. We can easily detect it by checking if HEAD is tagged with release version. Same about master - if HEAD is merge request generated by GitHub, and there are no local changes, we can be pretty sure it's master package.

utils/pckgen
```diff
@@ -0,0 +1,559 @@
...
+        VERSION="$VERSION_LONG"
+        RELEASE="devel.${LAST_COMMIT_TIMESTAMP}"
+    fi
+}
+
+create_version_file() {
+    echo "--- Creating version file"
+    if [ -f "$VERSION_FILE" ] && [ ! $FORCE ]; then
+        echo "---   Version file already exists (use --force to overwrite)"
+    else
```
robertbaldyga: Overwriting version file **really** sounds like a good idea. We want the version file to be always consistent with actual version of sources. Same about entire metadata directory content.

utils/pckgen
rafalste: This check is here to prevent building tar package if it was already built by e.g. 'generate_rpm()'. But I can see it now being very unintuitive. I'll find a different solution.

utils/pckgen
```diff
@@ -0,0 +1,454 @@
...
+# through this script all you need to do is:
+# - add 'case' command to set variable 'GENERATE_<NAME>=1'
+# - add function 'generate_<name>' to handle the task
+# - add 'DEPENDENCIES_<NAME>' array if needed
+# - update help!
+
+
+NAME="open-cas-linux"
+# TODO: change 'supported from' to 20.3
+SUPPORTED_FROM_VERSION="20.1"
```
arutk: I suggest to :
1. Change current version in CAS_VERSION_GEN to 20.3
2. Change this variable to 20.3

utils/pckgen
```diff
@@ -0,0 +1,454 @@
...
+        source "$VERSION_FILE" &> /dev/null
+    else
+        error "Couldn't obtain CAS version - no git tree nor version file within given sources"
+    fi
+
+    VERSION_LONG="${CAS_VERSION_MAIN}.${CAS_VERSION_MAJOR}.${CAS_VERSION_MINOR}.${CAS_VERSION_BUILD}"
+
+    if [ $(git -C "$SOURCES_DIR" tag --points-at HEAD) ]; then
+        VERSION="$VERSION_SHORT"
+        RELEASE="release"
```
arutk: My understanding is that the resulting "release" package will be identified as a lower version than preceding "master" packages. I believe this is not desired. It would be better to add a full version number in release package. @robertbaldyga do you agree?

robertbaldyga: Yep. Additionally one of main objectives was to have one unified version format, which IMO should be `VERSION_LONG`. Thus we want to use it also for release builds.

utils/pckgen
```diff
@@ -0,0 +1,450 @@
...
+        LAST_COMMIT_HASH_ABBR=$(git -C "$SOURCES_DIR" log -1 --pretty=format:%h)
+        LAST_COMMIT_DATE=$(git -C "$SOURCES_DIR" log -1 --pretty=format:%ci |\
+                           sed "s/ /T/" | sed "s/ //" | sed "s/00$/:00/")
+        LAST_COMMIT_TIMESTAMP=$(git -C "$SOURCES_DIR" log -1 --pretty=format:%ct)
+        for SUBMOD in ${SUBMODULES[@]}; do
+            LAST_SUB_COMMIT_HASHES+=($(git -C "$SOURCES_DIR/$SUBMOD" log -1 --pretty=format:%H))
+            LAST_SUB_COMMIT_HASHES_ABBR+=($(git -C "$SOURCES_DIR/$SUBMOD" log -1 --pretty=format:%h))
+        done
+    elif [ -f "$VERSION_FILE" ]; then
+        source "$VERSION_FILE" &> /dev/null
```
arutk: I'm not sure if we should support this case. This would be used to, for example, build rpm from an extracted tar archive created by this script? Do you think anyone would use it?

robertbaldyga: As *.tar.gz and *.zip archives are our official way of distributing sources, user should be able to produce all supported outcomes of Makefile using those packages.

PR #349, opened 2020-03-10
Allow activity threshold and wake up time in ALRU set to 0s

imjfckm:
Update OCF with change to ALRU allowing wider range of parameters + test to make sure it works allright.

### DIFF
--- a/test/functional/api/cas/cli.py
+++ b/test/functional/api/cas/cli.py
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2019 Intel Corporation
+# Copyright(c) 2019-2020 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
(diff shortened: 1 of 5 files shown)


test/functional/tests/cleaning/test_alru.py
```diff
@@ -0,0 +1,118 @@
+#
+# Copyright(c) 2019 Intel Corporation
```
Ostrokrzew: 2020

Ostrokrzew: 2020

test/functional/api/cas/cli.py
```diff
@@ -215,13 +215,14 @@ def set_param_cleaning_alru_cmd(cache_id: str, wake_up: str, staleness_time: str
                                 shortcut: bool = False):
     add_param = ""
     if wake_up is not None:
-        add_param += (" -w " if shortcut else " --wake-up ") + wake_up
+        add_param += (" -w " if shortcut else " --wake-up ") + str(wake_up)
```
arutk: Looks like you've change this function to accept non-string values of "wake_up" and other parameters. If that's your intention, please remove "str" type specification from argument list.

test/functional/tests/cleaning/test_alru.py
```diff
@@ -0,0 +1,118 @@
...
+        dirty_now = cache.get_statistics().usage_stats.dirty
+
+        if dirty_before > dirty_now:
+            TestRun.fail(
+                f"Cleaning has run, while it shouldn't"
+                " (dirty down from {dirty_before} to {dirty_now}"
+            )
+
+    with TestRun.step("Set 0 idle time and wake up time for ALRU"):
+        cache.set_params_alru(FlushParametersAlru(activity_threshold=Time(0)))
```
arutk: You declare you set both idle time and wake up time, but it looks like you just set one of them.

test/functional/tests/cleaning/test_alru.py
```diff
@@ -0,0 +1,118 @@
...
+    Udev.disable()
+    cache = casadm.start_cache(cache_dev.partitions[0], force=True, cache_mode=CacheMode.WB)
+    core = cache.add_core(core_dev.partitions[0])
+    cache.set_seq_cutoff_policy(SeqCutOffPolicy.never)
+    cache.set_cleaning_policy(CleaningPolicy.alru)
+    cache.set_params_alru(
+        FlushParametersAlru(
+            activity_threshold=Time(seconds=100),
+            staleness_time=Time(seconds=1),
+            wake_up_time=Time(0),
```
arutk: Why start with wake_up_time == 0 ?

test/functional/tests/cleaning/test_alru.py
imjfckm: Fixed

test/functional/api/cas/cli.py
imjfckm: True that! Removed type hints here

test/functional/tests/cleaning/test_alru.py
imjfckm: Yup, I must've messed it up while moving steps around. I think that 0 wake up time here is still reasonable to speedup the test - so I moved it to here from prepare.

test/functional/tests/cleaning/test_alru.py
imjfckm: No reason! Moved to more apropriate place ;)

PR #355, opened 2020-03-17
Async stop compl

mmichal10:


test/smoke_test/cas_config
```diff
@@ -10,8 +10,8 @@
 # Default core and cache devices - note that we require whole devices, not partitions.
 # Those are now empty to prevent from using default values by mistake, which could lead to
 # destroying partitions and data. If you want to setup those, use cas_local_config
-export CORE_DEVICE=""
-export CACHE_DEVICE=""
+export CORE_DEVICE="/dev/sdb"
+export CACHE_DEVICE="/dev/nvme0n1"
```
Ostrokrzew: You added values, but now description is confusing.

modules/cas_cache/layer_cache_management.c
```diff
@@ -448,39 +465,44 @@ static void _cache_mngt_cache_stop_rollback_complete(ocf_cache_t cache,
...
-	struct _cache_mngt_async_context *context;
+	struct _cache_mngt_stop_context *context;
 	int result = 0;
 
 	context = env_malloc(sizeof(*context), GFP_KERNEL);
 	if (!context)
 		return -ENOMEM;
 
-	_cache_mngt_async_context_init(context);
+	context->finish_thread = kthread_create(exit_instance_finish, context,
```
arutk: I think this should also be moved somewhere to init phase.  This function is used in initialization error handling path (see err_classifier label) - and in case of _cache_mngt_cache_stop_sync error we leak resources.

Looks to me that the entire _cache_mngt_stop_context  together with context->finish_thread should be allocated on init phase and stored somewhere in cache priv.

modules/cas_cache/layer_cache_management.c
```diff
@@ -406,20 +422,24 @@ struct _cache_mngt_attach_context {
...
+	if (kthread_should_stop())
+		return 0;
+
 	if (ctx->cls_inited)
 		cas_cls_deinit(cache);
 
 	if (ctx->priv_inited) {
 		cache_priv = ocf_cache_get_priv(cache);
+		kthread_stop(cache_priv->stop_context->finish_thread);
+		kfree(cache_priv->stop_context);
```
arutk: cache_priv->stop_context is initialized inside _cache_mngt_cache_priv_init(), so it would make sense to deinitialize it inside _cache_mngt_cache_priv_deinit().

modules/cas_cache/layer_cache_management.c
```diff
@@ -406,15 +425,17 @@ struct _cache_mngt_attach_context {
...
 {
 	struct cache_priv *cache_priv;
 	ocf_queue_t mngt_queue = NULL;
-	struct _cache_mngt_attach_context *ctx =
-		container_of(work, struct _cache_mngt_attach_context, work);
+	struct _cache_mngt_attach_context *ctx = data;
 	ocf_cache_t cache = ctx->cache;
 	int result;
 
+	if (kthread_should_stop())
```
imjfckm: Shouldn't you put the module in this path as well?

mmichal10: This condition is true only if synchronous `kthread_stop()` was called. In this case it is caller responsibility to put module.

PR #362, opened 2020-03-19
Fix core removing

mmichal10:
Patch fixes scenario when core-remove command hangs when called on inactive core.

Before core can be removed, it has to flushed. To flush it, read management lock has to acquired on cache. Command hanged because in case of trying to flush inactive core `_cache_mngt_remove_core_prepare()` returned without freeing lock.

Splitting `_cache_mngt_remove_core_prepare()` into two separate functions - one only for flushing core and second for destroying exported object made all core removing scenarios easier to handle.

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -538,13 +538,23 @@ static uint16_t find_free_core_id(uint64_t *bitmap)
 	return ret;
 }
 
-static void mark_core_id_used(uint64_t *bitmap, uint16_t core_id)
+static void mark_core_id_used(ocf_cache_t cache, uint16_t core_id)
 {
+	uint64_t *bitmap;
+	struct cache_priv *cache_priv = ocf_cache_get_priv(cache);
+
+	bitmap = cache_priv->core_id_bitmap;
+
 	set_bit(core_id, (unsigned long *)bitmap);
 }
 
-static void mark_core_id_free(uint64_t *bitmap, uint16_t core_id)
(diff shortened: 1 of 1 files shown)


modules/cas_cache/layer_cache_management.c
```diff
@@ -1248,26 +1252,22 @@ int cache_mngt_remove_core_from_cache(struct kcas_remove_core *cmd)
 	 * destroyed, instead of trying rolling this back we rather detach core
 	 * and then inform user about error.
 	 */
-	result = _cache_mngt_remove_core_prepare(cache, core, cmd, true);
-	if (result == -KCAS_ERR_REMOVED_DIRTY) {
-		flush_result = result;
-		result = 0;
-	} else if (result) {
+	prepare_result = _cache_mngt_remove_core_prepare(cache, core, cmd);
+	if (result && result != -KCAS_ERR_REMOVED_DIRTY)
```
micrakow: I think prepare_result should be checked here

modules/cas_cache/layer_cache_management.c
```diff
@@ -1150,41 +1171,54 @@ int _cache_mngt_remove_core_prepare(ocf_cache_t cache, ocf_core_t core,
...
-		}
-	}
+	if (!cmd->force_no_flush)
+		result = _cache_mngt_core_flush_uninterruptible(core);
 
-	if (flush_result)
-		result = destroy ? -KCAS_ERR_REMOVED_DIRTY : flush_result;
-
-	return result;
+	return result ? -KCAS_ERR_REMOVED_DIRTY : 0;
```
arutk: looks like 'result' is potentially uninitialized

PR #363, opened 2020-03-23
Show progress bar during changing cache mode

mmichal10:


modules/cas_cache/layer_cache_management.c
```diff
@@ -1992,39 +2012,45 @@ int cache_mngt_set_cache_mode(const char *cache_name, size_t name_len,
...
+	if (result)
+		goto unlock;
+
+	result = ocf_mngt_cache_set_mode(cache, mode);
+	if (result)
+		goto unlock;
 
 	result = _cache_mngt_save_sync(cache);
 	if (result)
 		ocf_mngt_cache_set_mode(cache, old_mode);
```
micrakow: Just wondering - do we maybe want to log this case or not?

modules/cas_cache/layer_cache_management.c
```diff
@@ -1992,39 +2012,45 @@ int cache_mngt_set_cache_mode(const char *cache_name, size_t name_len,
...
 	if (result)
 		return result;
 
-	result = _cache_mngt_lock_sync(cache);
-	if (result) {
-		ocf_mngt_cache_put(cache);
-		return result;
+	old_mode = ocf_cache_get_mode(cache);
+	if (old_mode == mode) {
+		printk(KERN_INFO "%s is in defined cache mode already", cache_name);
```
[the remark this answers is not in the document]
mmichal10: I agree, however afaik there is no function to translate cache mode id to cache mode name. Since previous cache mode can always be retrieved from logs, I would leave this log as it is.

modules/cas_cache/layer_cache_management.c
```diff
@@ -1999,32 +2019,43 @@ int cache_mngt_set_cache_mode(const char *cache_name, size_t name_len,
...
```
arutk: would be nice to repeat this check after acquiring management write lock - just to make sure cache mode did not change in between

modules/cas_cache/layer_cache_management.c
```diff
@@ -245,6 +245,13 @@ static void _cache_mngt_cache_flush_complete(ocf_cache_t cache, void *priv,
 		kfree(context);
 }
 
+/*
+ * Possible return values:
+ * 0 - completion was called and operation succeded
+ * -KCAS_ERR_WAITING_INTERRUPTED - operation was canceled, caller must
+ *		propagate error, completion will be called asynchronously
+ * oter values - completion was called and operation failed
```
robertbaldyga: s/oter/other

modules/cas_cache/layer_cache_management.c
```diff
@@ -291,6 +301,13 @@ static void _cache_mngt_core_flush_complete(ocf_core_t core, void *priv,
...
```
robertbaldyga: s/oter/other

PR #369, opened 2020-03-27
Before stop wait for start to be finished

mmichal10:
To prevent cache being stopped before async start finialize has finished, stop
routine have to check if worker responsible for finalizing is not running
anymore.

To create link between start finalize and stop functions, work_struct had to be
moved from attach_context to cache_priv.

### DIFF
--- a/modules/cas_cache/cas_cache.h
+++ b/modules/cas_cache/cas_cache.h
@@ -58,9 +58,11 @@ struct cas_classifier;
 struct cache_priv {
 	uint64_t core_id_bitmap[DIV_ROUND_UP(OCF_CORE_MAX, 8*sizeof(uint64_t))];
 	struct cas_classifier *classifier;
+	struct work_struct start_worker;
 	struct _cache_mngt_stop_context *stop_context;
 	atomic_t flush_interrupt_enabled;
 	ocf_queue_t mngt_queue;
+	void *attach_context;
 	ocf_queue_t io_queues[];
 };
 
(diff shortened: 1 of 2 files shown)


modules/cas_cache/cas_cache.h
```diff
@@ -58,9 +58,11 @@ struct cas_classifier;
 struct cache_priv {
 	uint64_t core_id_bitmap[DIV_ROUND_UP(OCF_CORE_MAX, 8*sizeof(uint64_t))];
 	struct cas_classifier *classifier;
+	struct work_struct start_worker;
 	struct _cache_mngt_stop_context *stop_context;
 	atomic_t flush_interrupt_enabled;
 	ocf_queue_t mngt_queue;
+	void *ctx;
```
arutk: please name this so that it's clear what kind of context is this

modules/cas_cache/layer_cache_management.c
```diff
@@ -1894,6 +1896,9 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
 	if (result != -KCAS_ERR_WAITING_INTERRUPTED)
 		kfree(context);
 
+	if (!result)
+		cache_priv->ctx = NULL;
```
arutk: is this needed?

mmichal10: Not needed, but at this point, `ctx` would point to freed memory, so it is I would call it nice to have.

PR #370, opened 2020-03-27
Fix git calls in package generator script

rafalste:
- make git command compatible with older git versions by removing -C option
- minor syntax improvements in the entire script

### DIFF
--- a/utils/pckgen
+++ b/utils/pckgen
@@ -6,7 +6,7 @@
 
 # To make additional packages creation possible
 # through this script all you need to do is:
-# - add 'case' command to set variable 'GENERATE_<NAME>=1'
+# - add 'case' command to set variable GENERATE_<NAME>="generate_<name>"
 # - add function 'generate_<name>' to handle the task
 # - add 'DEPENDENCIES_<NAME>' array if needed
 # - update help!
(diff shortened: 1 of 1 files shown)


utils/pckgen
```diff
@@ -151,28 +152,28 @@ check_version() {
...
-    if [ $(git -C "$SOURCES_DIR" tag --points-at HEAD) ]; then
+    if [ $(cd "$SOURCES_DIR" && git tag --points-at HEAD) ]; then
         RELEASE="release"
-    elif [ $(git -C "$SOURCES_DIR" log -1 --pretty=format:%H)\
-           == $(git -C "$SOURCES_DIR" log -1 --merges --pretty=format:%H) ]; then
+    elif [ $(cd "$SOURCES_DIR" && git log -1 --pretty=format:%H)\
+           == $(cd "$SOURCES_DIR" && git log -1 --merges --pretty=format:%H) ]; then
         RELEASE="master"
     else
         RELEASE="devel.${LAST_COMMIT_TIMESTAMP}"
```
robertbaldyga: This info should also be part of version file. In case there is no git repository (e.g. sources were extracted from archive) we will always fallback to "devel", which in most cases will be wrong, as zip/tar repositories will mostly be used in case of release versions.

PR #373, opened 2020-03-30
Improve _env_allocator_item fields alignment

arutk:
This change assures that data array is always optimally aligned.
Additionally item boolean flags are put into the same uint32
as cpu number in order to save space for future additions.

This change fixes ~50% performance degradation introduced by
commit d822a1d1.

### DIFF
--- a/modules/cas_cache/ocf_env.c
+++ b/modules/cas_cache/ocf_env.c
@@ -34,10 +34,10 @@ static inline size_t env_allocator_align(size_t size)
 }
 
 struct _env_allocator_item {
-	uint32_t cpu;
-	uint8_t from_rpool : 1;
-	uint8_t used : 1;
-	char data[];
+	uint32_t cpu : order_base_2(NR_CPUS);
+	uint32_t from_rpool : 1;
+	uint32_t used : 1;
+	char data[] __attribute__ ((aligned (__alignof__(uint64_t))));
 };
 
 void *env_allocator_new(env_allocator *allocator)


modules/cas_cache/ocf_env.c
```diff
@@ -102,6 +104,8 @@ env_allocator *env_allocator_create(uint32_t size, const char *name)
 		goto err;
 	}
 
+	BUG_ON(num_online_cpus() >= 1 << CAS_RPOOL_MAX_CPU_BITS);
```
robertbaldyga: This should be `>`, not `>=`. We also may want to perform this sanity check at the very beginning of this function. Nevertheless, taking into account that you just check if machine has more than over 3 billion (!) CPUs, it seems to be well overdone.

imjfckm: <insert Bill Gates 640K quote>

arutk: This BUG_ON is in case someone adds extra fields to this bitfield and lowers CAS_RPOOL_MAX_CPU_BITS

arutk: Also I think > is ok

arutk: This function is called only once during initialization, so impact is negligible

arutk: I neeed to put this check after variables initialization. And this check for NULL allocator looks good close to allocator initialization. I prefer it this way, but I can change this if you insist.

PR #374, opened 2020-03-31
Stop cache threads synchronously

mmichal10:
If adapter's callback to stop queue is asynchronous there is possible race conditions - queue memory might be freed before thread's loop is stopped.  This leads to dereferencing freed memory. 

Asynchronous stopping queue was remedy for stopping management queue from cache stop/rollback, however since putting queue was moved to separate thread/worker this problem no longer exists.

### DIFF
--- a/modules/cas_cache/threads.c
+++ b/modules/cas_cache/threads.c
@@ -51,10 +51,8 @@ static int _cas_io_queue_thread(void *data)
 	/* If we get here, then thread was signalled to terminate.
 	 * So, let's free memory and exit.
 	 */
-	wait_for_completion(&info->compl);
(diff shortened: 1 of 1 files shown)


modules/cas_cache/threads.c
```diff
@@ -207,11 +198,13 @@ static void _cas_start_thread(struct cas_thread_info *info)
 static void _cas_stop_thread(struct cas_thread_info *info)
 {
 	if (info && info->thread) {
-		reinit_completion(&info->compl);
+		init_completion(&info->compl);
```
arutk: reinit_completion looks better

PR #379, opened 2020-04-02
Weak modules

Ostrokrzew:
Tests for loading OpenCAS on startup after OS boot to different subversion or version of kernel.
Needs https://github.com/Open-CAS/test-framework/pull/145

Signed-off-by: Slawomir Jankowski slawomir.jankowski@intel.com


test/functional/tests/startup/test_startup_switch_kernels.py
```diff
@@ -0,0 +1,257 @@
...
+    with TestRun.step("Create filesystem, add core and mount it."):
+        core_part.create_filesystem(Filesystem.ext4)
+        core = cache.add_core(core_part)
+        core.mount(mountpoint)
+
+    with TestRun.step("Create test file and calculate md5 checksum."):
+        test_file = fs_utils.create_random_test_file(filepath, Size(16, Unit.MebiByte))
+        md5_before = test_file.md5sum()
+
+    with TestRun.step("Add mountpoint fstab and create opencas.conf."):
```
katlapinka, line 57: add mountpoint to fstab

test/functional/tests/startup/test_startup_switch_kernels.py
```diff
@@ -0,0 +1,257 @@
...
+        casadm.stop_all_caches()
+
+
+@pytest.mark.remote_only
+@pytest.mark.require_incompatible_kernel
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_cas_startup_incompatible_kernel():
+    """
+    title: Test for starting CAS on system startup after reboot to compatible kernel.
```
katlapinka, line 112: incompatibile

test/functional/tests/startup/test_startup_switch_kernels.py
```diff
@@ -0,0 +1,257 @@
...
+
+    with TestRun.step("Check if cache is started."):
+        caches = list(get_caches())
+        if len(caches) != 1:
+            pytest.xfail(f"Should not can list caches.")
+
+    with TestRun.step("Check if core is added."):
+        cores = list(get_cores(cache.cache_id))
+        if len(cores) != 1:
+            pytest.xfail(f"Should not can list cores.")
```
katlapinka, line 160: "should not can" - it is not a correct syntax. Please fix in in the whole patch.

test/functional/api/kernel.py
```diff
@@ -0,0 +1,121 @@
+from core.test_run_utils import TestRun
```
mmichal10, line 1: I wonder if `test-framework/test_utils/os_utils.py` wouldn't be a better place for kernel utils. There is no CAS specific operations here.

test/functional/tests/conftest.py
```diff
@@ -180,3 +181,24 @@ def base_prepare(item):
...
+
+def add_marks(config):
+    config.addinivalue_line(
+        "markers",
+        "require_compatible_kernel: require minimum two compatible kernel versions"
+    )
+    config.addinivalue_line(
+        "markers",
+        "require_incompatible_kernel: require minimum two incompatible kernel versions"
+    )
```
mmichal10, line 194: This also isn't CAS specific and can be moved to TF

test/functional/tests/startup/test_startup_switch_kernels.py
```diff
@@ -102,3 +102,79 @@ def test_cas_startup_compatible_kernel():
...
+        previous_kernel = check_current_kernel()
+        switch_kernel(compatible=False)
+
+    with TestRun.step("Reboot."):
+        TestRun.executor.reboot()
+
+    with TestRun.step("Check if cache is started."):
+        caches = list(get_caches())
+        if len(caches) != 1:
+            pytest.xfail(f"Should not list caches.")
```
mmichal10, line 154: Please use `TestRun.fail()`

test/functional/tests/startup/test_startup_switch_kernels.py
```diff
@@ -102,3 +102,79 @@ def test_cas_startup_compatible_kernel():
...
+            pytest.xfail(f"Should not list cores.")
+
+    with TestRun.step("Check if core is mounted."):
+        if not core.is_mounted():
+            pytest.xfail("Should not be mounted.")
+
+    with TestRun.step("Check if md5 checksum matches."):
+        md5_after = test_file.md5sum()
+        if md5_before != md5_after:
+            pytest.xfail("Md5 checksum should mismatch.")
```
mmichal10, line 168: I think conditions and error messages are kind of confusing. I.e. you check if number of started cache instances is not different than 1 but you print message like you would expect it to be zero.

test/functional/tests/startup/test_startup_switch_kernels.py
Ostrokrzew: It is expected for this step to fail.

test/functional/tests/conftest.py
Ostrokrzew: > Robert Baldyga
Those should probably be moved to OCL repository. TF doesn't provide any functionalities related to kernel version, so TF user cannot benefit from these marks in any way.

in comment to closed PR: https://github.com/Open-CAS/test-framework/pull/145/files

test/functional/tests/startup/test_startup_switch_kernels.py
Ostrokrzew: This and below steps either could pass or fail. It is expected that on incompatible kernels, OpenCAS may not load.

PR #385, opened 2020-04-06
Return CAS internal error if removing core fails

mmichal10:
When failed to open exported object exclusively, `EBUSY` is returned. If casadm
retrives such error code it misinterpres it and retries to call ioctl intead of
returing error and informative message to user.

To avoid pointless calling the same ioctl over and over again, adapter should
return `-KCAS_ERR_DEV_PENDING` when faild to destory exported object because it
was unable to open it exclusively.


mmichal10:
retest this please

KamilLepek:
retest this please

PR #386, opened 2020-04-06
Update version generating script

Ostrokrzew:
Update script, which generate modules' versions.
Modify CLI test for 'help' command.
Add test for 'version' command based on updated 'help' test.

### DIFF
--- a/modules/CAS_VERSION_GEN
+++ b/modules/CAS_VERSION_GEN
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
+SOURCES_DIR="$(pwd)/.."
 VER_FILE=CAS_VERSION
 
 if [ -f ${VER_FILE} ]; then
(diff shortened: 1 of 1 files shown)


modules/CAS_VERSION_GEN
```diff
@@ -29,3 +30,76 @@ echo "CAS_VERSION_MINOR=${CAS_VERSION_MINOR}" >> ${VER_FILE}
...
+    else
+        CAS_RELEASE="devel.${LAST_COMMIT_TIMESTAMP}"
+    fi
+elif [ -f "$VERSION_FILE" ]; then
+    source "$VERSION_FILE" &>/dev/null
+else
+    echo -e "Couldn't obtain CAS version - no git tree nor version file within given sources"
+fi
+
+CAS_VERSION="${CAS_VERSION_MAIN}.${CAS_VERSION_MAJOR}.${CAS_VERSION_MINOR}.${CAS_VERSION_BUILD}"
```
Ostrokrzew, line 90: It's nearly the same what's in *make package* script, but last member is 6 digits long, not four and minor version member also have two digits like major version. This part uses previously generated (at the beginning) **version file** which is afterwards overwritten by newer version's records.

PR #388, opened 2020-04-07
Rebuild test for *help* command

Ostrokrzew:
Add file with *help* messages.
Test all *help* call cases, even wrong.

based on https://github.com/Open-CAS/open-cas-linux/pull/386


test/functional/api/cas/cli_help_messages.py
```diff
@@ -0,0 +1,276 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+casadm_help = [
+    r"Cache Acceleration Software Linux",
+    r"Usage: casadm \<command\> \[option\.\.\.\]",
```
rafalste, line 8: I'm not sure if you need to (or even can) escape characters with backslash in raw strings. If you tested it and it's ok, then I'm backing off. ;) The rest looks good to me.

test/functional/tests/cli/test_cli_help_and_version.py
```diff
@@ -98,3 +98,20 @@ def test_cli_help(shortcut):
...
+    """
+    title: Test for 'version' command.
+    description: Test if version displays.
+    pass_criteria:
+      - Proper OCL's components names displays in table with its versions.
+    """
+    TestRun.LOGGER.info("Check OCL's version.")
+    output = casadm.print_version(shortcut=shortcut).stdout
+    TestRun.LOGGER.info(output)
+    if ("CAS Cache Kernel Module" not in output
```
micrakow, line 114: Just wondering - maybe we do want to check if version is displayed in expected format (e.g. "20.06.00.00000518")?

Ostrokrzew: Good point, I will prepare a change and push it when unified versioning becomes reality.

test/functional/api/cas/cli_help_messages.py
Ostrokrzew: Yeah, it is required by regex.

PR #389, opened 2020-04-07
Don't retry removing core when its busy

Ostrokrzew:

### DIFF
--- a/casadm/cas_lib_utils.c
+++ b/casadm/cas_lib_utils.c
@@ -469,12 +469,23 @@ int run_ioctl_interruptible(int fd, int command, void *cmd,
 }
 
 /*
- * @brief ioctl wrapper that retries ioctl attempts within one second timeouts
+ * @brief ioctl wrapper
  * @param[in] fd as for IOCTL(2)
  * @param[in] command as for IOCTL(2)
  * @param[inout] cmd_info as for IOCTL(2)
  */
 int run_ioctl(int fd, int command, void *cmd)
+{
+	return ioctl(fd, command, cmd);
+}
+
+/*
+ * @brief ioctl wrapper that retries ioctl attempts within one second timeouts
+ * @param[in] fd as for IOCTL(2)
+ * @param[in] command as for IOCTL(2)
+ * @param[inout] cmd_info as for IOCTL(2)
+ */
+int run_ioctl_retry(int fd, int command, void *cmd)
 {
 	int i, ret;
 	struct timespec timeout = {
--- a/casadm/cas_lib_utils.c
+++ b/casadm/cas_lib_utils.c
@@ -433,9 +433,10 @@ void *print_command_progress(void *th_arg)
  * Catch SIGINT signal.
  * @param friendly_name name of management operation that shall
  * be displayed in command prompt
+ * @param retry decide if ioctl attepmts should retry
  */
-int run_ioctl_interruptible(int fd, int command, void *cmd,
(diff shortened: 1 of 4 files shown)


casadm/cas_lib_utils.c
```diff
@@ -436,6 +436,22 @@ void *print_command_progress(void *th_arg)
...
+}
+
+/*
+ * Run ioctl in a way that displays progressbar (if flushing operation takes longer)
+ * Catch SIGINT signal.
+ * @param friendly_name name of management operation that shall
+ * be displayed in command prompt
+ * @param retry decide if ioctl attepmts should retry
+ */
+int run_ioctl_interruptible_retry_option(int fd, int command, void *cmd,
```
robertbaldyga, line 453: I'd make it separate functions with no additional arg.
Those running ioclt with default retry:
```
run_ioclt_retry()
run_ioclt_interruptible_retry()
```
and those running ioctl with no retry:
```
run_ioclt()
run_ioclt_interruptible()
```
That would be much easier to read.

casadm/cas_lib_utils.h
```diff
@@ -28,8 +28,13 @@ struct progress_status {
 void init_progress_bar(struct progress_status *ps);
 void print_progress_bar_or_indicator(float prog, struct progress_status *ps);
 int run_ioctl(int fd, int command, void *cmd);
+int run_ioctl_retry(int fd, int command, void *cmd);
 int run_ioctl_interruptible(int fd, int command, void *cmd,
 		char *friendly_name, int cache_id, int core_id);
+int run_ioctl_interruptible_retry(int fd, int command, void *cmd,
+		char *friendly_name, int cache_id, int core_id);
+int run_ioctl_interruptible_retry_option(int fd, int command, void *cmd,
+		char *friendly_name, int cache_id, int core_id, bool retry);
```
robertbaldyga, line 37: These should be static and removed from header. They are helper functions and they should be accessible only locally.

casadm/cas_lib_utils.c
```diff
@@ -474,6 +475,37 @@ int run_ioctl_interruptible(int fd, int command, void *cmd,
...
+ * @param friendly_name name of management operation that shall
+ * be displayed in command prompt
+ */
+int run_ioctl_interruptible(int fd, int command, void *cmd,
+		char *friendly_name, int cache_id, int core_id)
+{
+	bool retry = false;
+
+	return run_ioctl_interruptible_retry_option(fd, command, cmd, friendly_name, 
+				cache_id, core_id, retry);
```
micrakow, line 490: can't just false be passed here instead of creating variable and passing it here?

casadm/cas_lib_utils.c
```diff
@@ -474,6 +475,37 @@ int run_ioctl_interruptible(int fd, int command, void *cmd,
...
+ * @param friendly_name name of management operation that shall
+ * be displayed in command prompt
+ */
+int run_ioctl_interruptible_retry(int fd, int command, void *cmd,
+		char *friendly_name, int cache_id, int core_id)
+{
+	bool retry = true;
+
+	return run_ioctl_interruptible_retry_option(fd, command, cmd, friendly_name, 
+				cache_id, core_id, retry);
```
micrakow, line 506: as above

casadm/cas_lib_utils.c
Ostrokrzew: Ok, I've done it just for readability.

PR #391, opened 2020-04-09
Add test for installing OpenCA from RPM package

Ostrokrzew:
Needs https://github.com/Open-CAS/test-framework/pull/147


test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,64 @@
...
+        title: Test for installing OpenCAS from RPM
+        description: |
+          Prepare SRPM for OpenCAS, build OpenCAS's RPM from SRPM,
+          install OpenCAS from RPM and then uninstall OpenCAS.
+        pass_criteria:
+          - OpenCAS installs successfully
+          - OpenCAS uninstalls successfully
+    """
+    with TestRun.step("Check package manager."):
+        install_command = TestRun.executor.check_install_command()
```
robertbaldyga, line 26: Why do we need this? Isn't basic `rpm` tool enough for this test purposes?

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,64 @@
...
+    with TestRun.step("Check if OpenCAS is installed correctly."):
+        sleep(1)
+        try:
+            casadm.print_version()
+            TestRun.LOGGER.info("OpenCAS is installed as expected.")
+        except Exception:
+            TestRun.fail("OpenCAS should be installed.")
+
+    with TestRun.step("Uninstall OpenCAS from RPM."):
+        TestRun.executor.run(f"{install_command.split()[0]} remove open-cas-linux* -y")
```
robertbaldyga, line 57: It looks like the install command needs to be guessed by framework, but the remove command can be hardcoded :)
Again, using plain `rpm` solves the problem. I wouldn't expect RPM to be installable on the system, where `rpm` tool is not present anyway.

test/functional/tests/conftest.py
```diff
@@ -156,25 +156,55 @@ def base_prepare(item):
...
+                    casadm.stop_all_caches()
+                    casadm.remove_all_detached_cores()
+                except Exception:
+                    pass  # TODO: Reboot DUT if test is executed remotely
+                installer.uninstall_opencas()
+
+            for disk in TestRun.dut.disks:
+                disk.umount_all_partitions()
+                if not create_partition_table(disk, PartitionTable.gpt):
+                    raise Exception(f"Failed to remove partitions from {disk}")
```
robertbaldyga, line 174: Apparently half of the code in both `if` and `else` blocks is exactly the same. Why not just move it before `if-else` block?

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,64 @@
...
+
+import pytest
+
+from api.cas import casadm
+from core.test_run import TestRun
+
+
+@pytest.mark.remote_only()
+@pytest.mark.without_install()
+def test_load_x_to_one_without_params():
```
robertbaldyga, line 15: Name of this test function could be a bit move adequate to test contents.

test/functional/api/cas/installer.py
```diff
@@ -14,20 +14,8 @@
...
-        delete=True)
-
-    TestRun.LOGGER.info("Building Open CAS")
-    output = TestRun.executor.run(
-        f"cd {TestRun.usr.working_dir} && "
-        "./configure && "
-        "make -j")
-    if output.exit_code != 0:
-        raise CmdException("Make command executed with nonzero status", output)
+    download_opencas()
```
rafalste, line 17: Maybe 'copy_opencas()' will be a better name for this function, considering what it is doing?

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,73 @@
...
+        output = TestRun.executor.run_expect_success(
+            f"make rpm -C {TestRun.usr.working_dir}").stdout.splitlines()
+        interesting_output = []
+        for line in output:
+            if "RPMS/x86_64" in line:
+                interesting_output.append(line)
+            elif "You can find your fresh packages in" in line:
+                interesting_output.append(line)
+        program = interesting_output[0].split("/")[-1]
+        module = interesting_output[1].split("/")[-1]
```
rafalste, line 47: You shouldn't rely on printed package order when determining which one's which. Instead try to match package name (and ideally also a version).

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,73 @@
...
+        path = interesting_output[2].split("'")[1]
+
+    with TestRun.step("Install OpenCAS from RPM."):
+        TestRun.executor.run(f"{install_command} {path}/{module} {ignore}")
+        TestRun.executor.run(f"{install_command} {path}/{program} {ignore}")
+
+    with TestRun.step("Check if OpenCAS is installed correctly."):
+        sleep(1)
+        try:
+            casadm.print_version()
```
rafalste, line 57: There should also be a check if CAS is installed from RPM and not from sources. You can do this by checking the exit status of `rpm -qa | grep open-cas-linux`

test/functional/tests/install/test_install_rpm.py
mmichal10, line 47: You could create utility for rpm in `api/cas` - future tests could use it.

Ostrokrzew: Thanks for Your feedback. It was very helpful in refactoring this code.

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,81 @@
...
+
+    with TestRun.step("Get names of packages to install."):
+        packages = TestRun.executor.run(f"ls -1 {path_to_packages}").stdout.splitlines()
+        # For now there should be only 2 packages, if there would be more packages,
+        # we'll need to add more statements. Splitting packages makes installation easier
+        for package in packages:
+            if "modules" in package:
+                module = package
+            else:
+                program = package
```
rafalste, line 54: If you don't want to be dependent on number of packages, you can:
- list files with their absolute paths:
`ls -1 {path_to_packages}/*`
- create a list to install:
`packages_to_install = ' '.join(packages)`
- and then simply install:
`rpm -i {packages_to_install}`

And to remove them:
- obtain packages names:
```
for package in packages:
    concatenate output of 'rpm -qp --qf %\{NAME\} {package}'
```
- pass that to `rpm -e`

test/functional/tests/conftest.py
```diff
@@ -164,19 +167,30 @@ def base_prepare(item):
...
         TestRun.LOGGER.add_build_info(f'Commit hash:')
         TestRun.LOGGER.add_build_info(f"{git.get_current_commit_hash()}")
         TestRun.LOGGER.add_build_info(f'Commit message:')
         TestRun.LOGGER.add_build_info(f'{git.get_current_commit_message()}')
+
+
+def add_marks(config):
+    config.addinivalue_line(
+        "markers",
+        "without_installation: don't install tested software and if already installed, uninstall it"
```
mmichal10, line 195: How about `uninstall_cas`? `without_installation` sounds like installation is skipped, but

test/functional/tests/conftest.py
```diff
@@ -165,19 +167,23 @@ def base_prepare(item):
...
             if not create_partition_table(disk, PartitionTable.gpt):
                 raise Exception(f"Failed to remove partitions from {disk}")
 
-        if get_force_param(item) and not TestRun.usr.already_updated:
+        if get_force_param(item) and not TestRun.usr.already_updated and not uninstall_cas:
             installer.rsync_opencas_sources()
             installer.reinstall_opencas()
-        elif not installer.check_if_installed():
+
+        elif not installer.check_if_installed() and not uninstall_cas:
```
mmichal10: I would move `if` from :171 here and remove newly added conditions.
    `if uninstall_cas:`
         ` installer.uninstall_opencas()`
  `  elif get_force_param(item) and not TestRun.usr.already_updated:`
          `....`

please keep in mind `uninstall_opencas()` may raise an exception.

test/functional/tests/conftest.py
```diff
@@ -178,11 +182,15 @@ def base_prepare(item):
 
         if get_force_param(item) and not TestRun.usr.already_updated and not uninstall_cas:
             installer.rsync_opencas_sources()
-            installer.reinstall_opencas()
+            installer.reinstall_opencas(
```
mmichal10, line 185: `installer.rsync_opencas_sources()` could be renamed to `installer.rsync_opencas()` and expect `copy_rpm` and `rpm_path` arguments

test/functional/config/example_dut_config.yml
```diff
@@ -13,6 +13,11 @@ type: "local"
 allow_disk_autoselect: False
 working_dir: "/tmp/open-cas-linux/"
 
+# This option is used only, when CLI flag '--rpm_install' is on. When flag is on
```
Deixx, line 16: change This to Below

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,61 @@
...
+        packages = rpm.get_packages_from_dir(path_to_packages)
+        rpm.install_packages(packages)
+
+    with TestRun.step("Load OpenCAS modules."):
+        output = os_utils.load_kernel_module(CasModule.cache.value)
+        if output.exit_code != 0:
+            TestRun.fail(f"Loading {CasModule.cache.value} module failed.")
+
+    with TestRun.step("Check if OpenCAS is installed correctly from RPM."):
+        sleep(1)
```
katlapinka, line 48: what is the purpose of this sleep?

test/functional/tests/conftest.py
```diff
@@ -113,15 +113,16 @@ def pytest_runtest_teardown():
...
     TestRun.configure(config)
 
 
 def pytest_addoption(parser):
     parser.addoption("--dut-config", action="store", default="None")
     parser.addoption("--log-path", action="store",
                      default=f"{os.path.join(os.path.dirname(__file__), '../results')}")
-    parser.addoption("--force-reinstall", action="store", default="False")
-    # TODO: investigate whether it is possible to pass the last param as bool
+    parser.addoption("--force-reinstall", action="store", default=False)
```
katlapinka, line 124: Did you verify if it works? I remember that we wanted to use bool type here but for some reason it didn't work and we decided to use string here.

test/functional/api/cas/installer.py
```diff
@@ -58,15 +64,26 @@ def install_opencas():
         TestRun.LOGGER.info(output.stdout)
 
 
-def set_up_opencas(version=None):
-    _clean_opencas_repo()
+def set_up_opencas(version=None, install_rpm: bool = False, path_to_rpm: str = None):
```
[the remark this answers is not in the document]
mmichal10: `install_rpm` should enable preparing rpm via installer if user doesn't want to provide one.

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,61 @@
...
+        if not installer.check_if_installed():
+            TestRun.fail("OpenCAS is not installed correctly.")
+        if not rpm.is_package_installed("open-cas-linux"):
+            TestRun.fail("OpenCAS is not installed from RPM.")
```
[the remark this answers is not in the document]
Ostrokrzew: I don't know if it is a good idea. This function returns only `True` or `False`, it doesn't return type of installation. Also if someone runs test on new DUT, she/ he cannot know if OpenCAS was preinstalled from sources or from RPM. Maybe another function could help, but what is the purpose? We just want to know if OpenCAS is installed in system, no matter how. The other function from `rpm` module is for checking how OpenCAS was installed and it is meaningful only for this test.

test/functional/tests/conftest.py
Ostrokrzew: I found this:
https://docs.pytest.org/en/latest/reference.html#parser
https://docs.python.org/3/library/argparse.html#action

test/functional/api/cas/installer.py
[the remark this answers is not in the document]
Ostrokrzew: You both were right. Fixed.

katlapinka:
retest this please

test/functional/tests/install/test_install_rpm.py
```diff
@@ -0,0 +1,56 @@
...
+        title: Test for installing OpenCAS from RPM
+        description: |
+          Check if OpenCAS could be installed from RPM and then uninstalled.
+        pass_criteria:
+          - OpenCAS installs successfully
+          - OpenCAS uninstalls successfully
+    """
+    with TestRun.step("Download and build OpenCAS."):
+        installer.rsync_opencas(install_rpm=True)
+        installer.set_up_opencas()
```
mmichal10, line 27: `set_up_opencas()` installs opencas from sources. Is this necessary during this test?

test/functional/tests/conftest.py
```diff
@@ -157,28 +164,44 @@ def base_prepare(item):
...
-            installer.set_up_opencas()
+        if uninstall_cas and opencas_installed:
+            RpmInstaller.rsync_opencas(install_rpm=install_rpm)
+            RpmInstaller.uninstall_opencas()
+        elif get_force_param(item) and not TestRun.usr.already_updated:
+            RpmInstaller.rsync_opencas(install_rpm=install_rpm)
+            RpmInstaller.reinstall_opencas(install_rpm=install_rpm)
+        elif not opencas_installed and not uninstall_cas:
+            RpmInstaller.rsync_opencas(install_rpm=install_rpm)
+            RpmInstaller.set_up_opencas(install_rpm=install_rpm)
```
mmichal10, line 194: If test has `uninstall` mark, but `force-reinstall` param is given, CAS will be installed anyway. IMO it sounds line misbehavior

Ostrokrzew: good point

katlapinka:
retest this please

test/functional/api/cas/installer.py
```diff
@@ -3,99 +3,150 @@
...
-    output = TestRun.executor.run("which casadm")
-    modules_loaded = os_utils.is_kernel_module_loaded(cas_module.CasModule.cache.value)
+        Installer.install_opencas()
 
-    if output.exit_code == 0 and modules_loaded:
-        TestRun.LOGGER.info("CAS is installed")
+    @staticmethod
+    def uninstall_opencas():
+        TestRun.LOGGER.info("Uninstalling Open CAS")
+        output = TestRun.executor.run(f"cd {TestRun.usr.working_dir} && make uninstall")
```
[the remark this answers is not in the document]
Ostrokrzew: Disagree. This method is named `uninstall_opencas`, not `check_if_installed_and_then_uninstall_opencas`. Checking that should be done before calling this method.

test/functional/api/cas/installer.py
```diff
@@ -3,99 +3,150 @@
...
+        if output.exit_code != 0:
+            return False
+        else:
+            TestRun.LOGGER.info(output.stdout)
+            return True
+
+
+class RpmInstaller(Installer):
+    @staticmethod
+    def rsync_opencas(install_rpm: bool = False):
```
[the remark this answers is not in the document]
Ostrokrzew: Nah, it's only a wrapper, which should give to user a possibility, not a coercion, to install CAS from RPM.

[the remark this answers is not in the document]
rafalste: But wouldn't it be more clear if this possibility will be given to the user before utilizing RpmInstaller class? As the name suggests RpmInstaller should be used to install RPM. It is a little confusing as it is right now. At least it was for me, when I was trying to understand this code.

test/functional/api/cas/installer.py
```diff
@@ -3,99 +3,150 @@
...
+        if install_rpm and TestRun.usr.rpm_dir is not None:
```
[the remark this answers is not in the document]
Ostrokrzew: I checked this statement and it's good. `is not None` refers only to `TestRun.usr.rpm_dir`.

test/functional/api/cas/installer.py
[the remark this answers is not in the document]
Ostrokrzew: You convinced me.

test/functional/api/cas/installer.py
```diff
@@ -3,99 +3,146 @@
...
+
+    @staticmethod
+    def _is_casadm_installed():
+        TestRun.LOGGER.info("Check if 'casadm' is properly installed.")
+        output = TestRun.executor.run("casadm -V")
+        if output.exit_code != 0:
+            return False
+        else:
+            TestRun.LOGGER.info(output.stdout)
+            return True
```
Deixx, line 105: 'which casadm' does the job, and we don't need output from casadm here

test/functional/tests/conftest.py
```diff
@@ -161,36 +165,57 @@ def get_force_param(item):
...
+        installer.rsync_opencas()
+
+        if uninstall_cas:
+            if opencas_installed:
+                RpmInstaller.uninstall_opencas()
+        elif get_force_param(item) and not TestRun.usr.already_updated:
             installer.reinstall_opencas()
-        elif not installer.check_if_installed():
-            installer.rsync_opencas_sources()
+        elif not opencas_installed and not uninstall_cas:
```
Deixx, line 207: uninstall_cas is always False here

test/functional/tests/conftest.py
```diff
@@ -161,36 +165,57 @@ def get_force_param(item):
...
-        if get_force_param(item) and not TestRun.usr.already_updated:
-            installer.rsync_opencas_sources()
+        # sources should be up-to-date on DUT even if installation is skipped
+        installer.rsync_opencas()
+
+        if uninstall_cas:
+            if opencas_installed:
+                RpmInstaller.uninstall_opencas()
+        elif get_force_param(item) and not TestRun.usr.already_updated:
             installer.reinstall_opencas()
```
Deixx, line 206: If RPM is installed and we want to reinstall from sources then installer.reinstall_opencas() will try to remove only source-based installation (and fail), and then will try to install from sources although RPM would still be installed.

Deixx (changes requested):

katlapinka:
Remove kedr test and update test for rpm installer

Deixx (commented):
test-framework patch lost

Issue #347, opened 2020-03-06
Build error message not informative


esmondchang:
No rule to make target 'inc' stop

mmichal10:
Hi @esmondchang! 

Could give us some more information? What is your OS, kernel version, compiler version?
Do you have ocf submodule in good version? Did you execute `git submodule update --init`?
Did the `confgiure` script return 0 status? 
Have you tried to clean open-cas-linux directory with `make distclean` (you can also try `make distclean && git clean -dxf`), and then configure and build it one more time?

mmichal10:
Hi @esmondchang!

Do you have ocf submodule in good version? Please execute `git submodule update --init`, and then try to build OCL.

Did the confgiure script return 0 status?
Have you tried to clean open-cas-linux directory with make distclean (you can also try make distclean && git clean -dxf), and then configure and build it one more time?

esmondchang:
Hi, 
you are right.  sorry for false alarm. i miss the step to git submodule update.  The configure didn't response any error till you get make error. If user forgot to have submodule update, shouldn't it be report during the configure?

Thanks...

robertbaldyga:
Print message about missing submodule in Makefile.

PR #393, opened 2020-04-14
Check for submodules in ./configure

rafalste:
Closes https://github.com/Open-CAS/open-cas-linux/issues/347

### DIFF
--- a/configure
+++ b/configure
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright(c) 2012-2019 Intel Corporation
+# Copyright(c) 2012-2020 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -25,6 +25,24 @@ check_util make
 check_util gcc
 check_util lsblk
 
+SCRIPTPATH=`dirname $0`
+SCRIPTPATH=`realpath $SCRIPTPATH`
+SUBMODULES=(
+    "ocf"
+    "test/functional/test-framework"
```
arutk, line 32: maybe get the list of submodules from git?

rafalste: But you may not have a git tree in your sources.

Ostrokrzew: It's true, but then why You print communicate "Do 'git submodule update --init' and try again!" below? If somebody will download zip or tarball, then he/ she would not be able to use 'git' commands from here before initiating this directory as git repo, but in this case, he/she would have submodules, so this check wouldn't result with false. This case is probable in case when somebody download OCL from GitHub with git, so I think Adam is right. You should take this list from file '.gitmodules'.

Ostrokrzew: I am convinced.

configure
```diff
@@ -25,6 +25,24 @@ check_util make
...
+    if ! ls -A "$SCRIPTPATH/$SUBMOD/"* &>/dev/null; then
+        SUBMODULES_MISSING+="'$SUBMOD' "
+    fi
+done
+if [ "$SUBMODULES_MISSING" ]; then
+    echo "Error: missing submodules: ${SUBMODULES_MISSING}" >&2
+    echo "Do 'git submodule update --init' and try again!" >&2
+    MISSING_TOOLS=1
+fi
+
```
mmichal10, line 45: Please convert spaces to tabs.

configure
```diff
@@ -25,6 +25,24 @@ check_util make
...
+)
+
+for SUBMOD in ${SUBMODULES[@]}; do
+    if ! ls -A "$SCRIPTPATH/$SUBMOD/"* &>/dev/null; then
+        SUBMODULES_MISSING+="'$SUBMOD' "
+    fi
+done
+if [ "$SUBMODULES_MISSING" ]; then
+    echo "Error: missing submodules: ${SUBMODULES_MISSING}" >&2
+    echo "Do 'git submodule update --init' and try again!" >&2
```
robertbaldyga, line 42: I'd replace "Do" with "Please run". It sounds much more polite IMO.

PR #409, opened 2020-05-07
Upgrade in flight for RPM

rafalste:

### DIFF
--- a/utils/upgrade_in_flight
+++ b/utils/upgrade_in_flight
@@ -1,0 +1,415 @@
+#!/bin/bash
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+
+THIS=$(basename "$0")
+CAS_NAME="open-cas-linux"
+UPGRADED_MODULE="cas_cache"
+DEPENDENCIES=(realpath basename dirname uname lsmod rmmod modprobe mktemp md5sum)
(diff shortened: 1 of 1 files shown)


imjfckm:
Isn't this something we'd like to do in utils/upgrade script? Seems to me like this is redundant and yet another utility serving more or less the same purpose. Or maybe remove utils/upgrade and implement this functionality here?

I think we should have some kind of discussion where do we want to go with this. @robertbaldyga  @arutk @mmichal10 ?

mmichal10:
@imjfckm @rafalste 

`utils/upgrade` script has few additional checks, which are missing in this script - check for empty corepool, check for missing cores. Those checks have to be added if we would like to merge @rafalste patch.

I would like you to keep in mind that our test API will also require utility for RPM. Do you think it be good idea to create one utility to be used by tests and by `utils/upgrade`? This could save us some time for maintenance in the future.

rafalste:
I could add those checks mentioned by @mmichal10 if they are necessary.

Regarding one utility which can be used by tests and by upgrade script, I think there might be some obstacles with this approach, but maybe someone more experienced can share his opinion.

imjfckm:
Overall I'm against creating new separate tool for RPM upgrades and think we should incorporate this functionality in already existing `utils/upgrade`. Unless it poses significant problems IMO that's the way we should go.

Deixx:

robertbaldyga:
We no longer support upgrade in flight.

PR #419, opened 2020-05-20
Update 'example_config' after changes in TF

Ostrokrzew:
The reason for this patch is https://github.com/Open-CAS/test-framework/pull/158

### DIFF
--- a/test/functional/config/example_dut_config.yml
+++ b/test/functional/config/example_dut_config.yml
@@ -7,6 +7,7 @@ type: "local"
 # password: "example_password"
 
 # env: "HTTPS_PROXY=myproxy.example:12345"
+# reboot_timeout: 1800  #timeout value in seconds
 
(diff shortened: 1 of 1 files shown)


test/functional/config/example_dut_config.yml
```diff
@@ -7,6 +7,7 @@ type: "local"
 # password: "example_password"
 
 # env: "HTTPS_PROXY=myproxy.example:12345"
+# reboot_timeout: "reboot_timeout_in_seconds"
```
katlapinka, line 10: "" can suggest that timeout should be a string, not an integer value
What do you think about removing "" or maybe set some default value to this variable and add some comment like:

#reboot_timeout: 1800 #timeout value in seconds

Ostrokrzew: That's a good idea, thanks.

PR #424, opened 2020-05-26
Improve installer

mmichal10:
Installing particular CAS version via test API will be used in upgrade-in-flight tests. 
Refactoring `test/functional/api/cas/installer.py` will allow to expand it with API for rpms

### DIFF
--- a/test/functional/api/cas/installer.py
+++ b/test/functional/api/cas/installer.py
@@ -58,13 +58,13 @@ def install_opencas():
         TestRun.LOGGER.info(output.stdout)
 
 
-def set_up_opencas():
+def set_up_opencas(version=None):
     rsync_opencas_sources()
 
     _clean_opencas_repo()
 
     if version:
-        git.checkout_cas_version()
+        git.checkout_cas_version(version)
 
     build_opencas()
 
@@ -84,10 +84,10 @@ def uninstall_opencas():
             raise CmdException("There was an error during uninstall process", output)
 
 
-def reinstall_opencas():
+def reinstall_opencas(version=None):
     if check_if_installed():
         uninstall_opencas()
-    set_up_opencas()
+    set_up_opencas(version)
 
 
 def check_if_installed():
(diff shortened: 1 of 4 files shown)


test/functional/api/cas/version.py
```diff
@@ -0,0 +1,74 @@
...
+# Copyright(c) 2019-2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import os
+
+from api.cas import git
+
+
+class CasVersion:
```
katlapinka, line 11: maybe you can extend existing python  Version class? This class has all comparison methods implemented etc

test/functional/tests/conftest.py
```diff
@@ -172,9 +172,9 @@ def base_prepare(item):
                 raise Exception(f"Failed to remove partitions from {disk}")
 
         if get_force_param(item) and not TestRun.usr.already_updated:
-            installer.reinstall_opencas()
+            installer.set_up_opencas()
```
rafalste, line 175: Shouldn't it be left as `reinstall_opencas()`?

test/functional/api/cas/installer.py
```diff
@@ -8,19 +8,46 @@
...
+def _clean_opencas_repo():
+    TestRun.LOGGER.info("Cleaning Open CAS repo")
+    output = TestRun.executor.run(
+        f"cd {TestRun.usr.working_dir} && "
+        "make distclean")
+    if output.exit_code != 0:
+        raise CmdException("make distclean command executed with nonzero status", output)
+
+
+def _checkout_version(version: CasVersion):
```
Deixx, line 35: it would make sense not to limit this to tags - checking out by hash will be necessary to test pre-release versions

test/functional/api/cas/installer.py
```diff
@@ -21,6 +21,18 @@ def install_opencas():
         exclude_list=["test/functional/results/"],
         delete=True)
 
+
+def _clean_opencas_repo():
+    TestRun.LOGGER.info("Cleaning Open CAS repo")
+    output = TestRun.executor.run(
```
micrakow, line 28: run_expect_success could be used

PR #428, opened 2020-06-01
Update test-framework and update example config

robertbaldyga:

### DIFF
--- a/test/functional/config/example_dut_config.yml
+++ b/test/functional/config/example_dut_config.yml
@@ -22,3 +22,10 @@ disks:
     serial: "DEF"
     type: "hdd"
     blocksize: 512
+
+plugins:
+  example_plugin:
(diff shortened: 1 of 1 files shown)


test/functional/config/example_dut_config.yml
```diff
@@ -22,3 +22,10 @@ disks:
     serial: "DEF"
     type: "hdd"
     blocksize: 512
+
+plugins:
+  example_plugin:
+    provided_by: "internal_plugins.example_plugin"
+    config:
+      property1: "value1"
+      property2: "value2"
```
mmichal10, line 31: Shouldn't it be commented be default?

robertbaldyga: Example plugin is noop so this shouldn't be a problem.

PR #432, opened 2020-06-05
Version unifying

rafalste:


utils/cas_version_gen
```diff
@@ -0,0 +1,86 @@
...
+    if [[ $(cd "$SOURCES_DIR" && git tag --points-at HEAD) ]]; then
+        CAS_RELEASE="release"
+    elif [[ $(cd "$SOURCES_DIR" && git log -1 --pretty=format:%H)\
+            == $(cd "$SOURCES_DIR" && git log -1 --merges --pretty=format:%H) ]]; then
+        CAS_RELEASE="master"
+    else
+        CAS_RELEASE="devel.${LAST_COMMIT_TIMESTAMP}"
+    fi
+
+    CAS_VERSION=$(printf "%02d.%02d.%01d.%04d-%s" $CAS_VERSION_MAIN $CAS_VERSION_MAJOR\
```
Ostrokrzew, line 41: You could save 2 digits for minor part of CAS version. Also I suggest more digits for release part. Now it's not needed, but in future number of commits could exceed 9999 records, but I think somebody else should decide.

PR #444, opened 2020-06-17
Performance test for 100% hit ratio

imjfckm:

### DIFF
--- a/test/functional/requirements.txt
+++ b/test/functional/requirements.txt
@@ -1,1 +1,2 @@
 attotime>=0.2.0
+schema==0.7.2
--- a/test/functional/tests/performance/test_100p_hits.py
+++ b/test/functional/tests/performance/test_100p_hits.py
@@ -1,0 +1,156 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+
+from api.cas import casadm
+from api.cas.cache_config import (
+    CacheMode,
+    CacheLineSize,
+    SeqCutOffPolicy,
+    CleaningPolicy,
+)
+from utils.performance import WorkloadParameter
+from core.test_run import TestRun
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import IoEngine, ReadWrite
+from test_utils.os_utils import Udev, set_wbt_lat, get_dut_cpu_physical_cores
+from test_utils.size import Size, Unit
+from test_utils.output import CmdException
+from storage_devices.disk import DiskTypeSet, DiskTypeLowerThan, DiskType
+
+
+@pytest.mark.performance()
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("queue_depth", [1, 4, 16, 64, 256])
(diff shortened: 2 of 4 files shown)


imjfckm:
I think, if we were to run this on NAND caches we'd also need some parametrization scheme, which would guarantee ascending order of qd*num_jobs.

But IMO we should think of custom parametrizations in broader context and as a separate effort - e.g. to reduce cache_mode*cls*cleaning_policy cases.

test/functional/tests/performance/test_100p_hits.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("queue_depth", [1, 4, 16, 64, 256])
+@pytest.mark.parametrize("numjobs", [1, 4, 16, 64, 256])
+@pytest.mark.parametrize("cache_line_size", CacheLineSize)
+def test_4k_100p_hit_reads_wt(queue_depth, numjobs, cache_line_size, perf_collector, request):
+    """
+        title: Test CAS performance in 100% Cache Hit scenario
+        decription: |
+          Characterize cache device with workload (parametrized by qd and jub number), and then run
```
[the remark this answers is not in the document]
Ostrokrzew: I mean 'jub'.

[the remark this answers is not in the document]
imjfckm: Fixed!

test/functional/utils/performance.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def insert_config_param(self, param, kind: ConfigParameter):
+        self.conf_params.insert_metric(param, kind)
+
+    def insert_config_from_cache(self, cache):
+        cache_config = {
+            "cache_mode": cache.get_cache_mode(),
+            "cache_line_size": cache.get_cache_line_size(),
+            "cleaning_policy": cache.get_cleaning_policy(),
+        }
```
mmichal10, line 150: I'm wondering if information about promotion policy and sequential cutoff would be wort to keep

test/functional/tests/performance/test_100p_hits.py
```diff
@@ -0,0 +1,155 @@
...
+    with TestRun.step("Run fio"):
+        cache_results = fio_cfg.run()[0]
+
+    perf_collector.insert_workload_param(numjobs, WorkloadParameter.NUM_JOBS)
+    perf_collector.insert_workload_param(queue_depth, WorkloadParameter.QUEUE_DEPTH)
+    perf_collector.insert_cache_metrics_from_fio_job(cache_dev_characteristics)
+    perf_collector.insert_exp_obj_metrics_from_fio_job(cache_results)
+
+
+def fill_cache(core, size):
```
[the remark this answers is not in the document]
imjfckm: Yes, I think the name is rather unfortunate. I think now it's more accurate

test/functional/utils/performance.py
imjfckm: For now we don't plan on testing performance with those settings. So in spirit of YAGNI I didn't include it.

test/functional/requirements.txt
```diff
@@ -1 +1,2 @@
 attotime>=0.2.0
+schema==0.7.2
```
katlapinka, line 2: Does it have to be equal or can it be ">="?

imjfckm: I think it could be. I just pinned it to this specific version, because that's the one I tested with.

PR #447, opened 2020-06-19
tests: add cache start trim test

micrakow:

### DIFF
--- a/test/functional/tests/trim/test_trim.py
+++ b/test/functional/tests/trim/test_trim.py
@@ -1,0 +1,105 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+import re
+
+from api.cas import casadm
+from core.test_run import TestRun
+from test_utils import os_utils
+from test_utils.size import Size, Unit
+from test_tools.fio.fio import Fio
+from test_tools.blktrace import BlkTrace, BlkTraceMask
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+def test_trim_start_discard():
+    """
+    title: Check discarding cache device at cache start
+    description: |
+       Create 2 partitions on trim-supporting device, write pattern to both partitions,
+       start blktrace against first one, start cache on first partition and check if discard
+       requests were sent at all and only to the first partition.
+    pass_criteria:
+      - Partition used for cache is discarded.
+      - Second partition is untouched - written pattern is preserved.
+    """
+    with TestRun.step("Clearing dmesg"):
+        TestRun.executor.run_expect_success("dmesg -C")
+
+    with TestRun.step("Preparing cache device"):
+        dev = TestRun.disks['cache']
+        dev.create_partitions([Size(500, Unit.MebiByte), Size(500, Unit.MebiByte)])
+        cas_part = dev.partitions[0]
+        non_cas_part = dev.partitions[1]
+
+    with TestRun.step("Writing different pattern on partitions"):
+        cas_fio = write_pattern(cas_part.system_path)
+        non_cas_fio = write_pattern(non_cas_part.system_path)
+        cas_fio.run()
+        non_cas_fio.run()
+
+    # TODO add blktracing for non-cas part
+    with TestRun.step("Starting blktrace against first (cache) partition"):
+        blktrace = BlkTrace(cas_part, BlkTraceMask.discard)
+        blktrace.start_monitoring()
+
+    with TestRun.step("Starting cache"):
+        cache = casadm.start_cache(cas_part, force=True)
+        metadata_size = get_metadata_size_from_dmesg()
+
+    with TestRun.step("Stop blktrace and check if discard requests were issued"):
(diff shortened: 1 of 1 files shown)


test/functional/tests/trim/test_trim.py
```diff
@@ -0,0 +1,105 @@
...
+
+    with TestRun.step("Starting cache"):
+        cache = casadm.start_cache(cas_part, force=True)
+        metadata_size = get_metadata_size_from_dmesg()
+
+    with TestRun.step("Stop blktrace and check if discard requests were issued"):
+        cache_reqs = blktrace.stop_monitoring()
+        cache_part_start = cas_part.begin
+
+        if len(cache_reqs) == 0:
```
mmichal10, line 60: Where did you triggered discard?

test/functional/tests/trim/test_trim.py
```diff
@@ -0,0 +1,105 @@
...
+        cas_part = dev.partitions[0]
+        non_cas_part = dev.partitions[1]
+
+    with TestRun.step("Writing different pattern on partitions"):
+        cas_fio = write_pattern(cas_part.system_path)
+        non_cas_fio = write_pattern(non_cas_part.system_path)
+        cas_fio.run()
+        non_cas_fio.run()
+
+    # TODO add blktracing for non-cas part
```
mmichal10, line 47: Wouldn't it be just copy-paste of blktracing cas part?

test/functional/tests/conftest.py
```diff
@@ -169,8 +169,7 @@ def base_prepare(item):
 
         for disk in TestRun.dut.disks:
             disk.umount_all_partitions()
-            if not create_partition_table(disk, PartitionTable.gpt):
-                raise Exception(f"Failed to remove partitions from {disk}")
+            create_partition_table(disk)
```
mmichal10, line 172: Why did you change it? You could add it as separate commit with explanation in commit message

test/functional/tests/trim/test_trim.py
```diff
@@ -0,0 +1,105 @@
...
+from core.test_run import TestRun
+from test_utils import os_utils
+from test_utils.size import Size, Unit
+from test_tools.fio.fio import Fio
+from test_tools.blktrace import BlkTrace, BlkTraceMask
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+
+
+#TODO Add mark for getting discard-supproting devcie
```
mmichal10, line 19: How about additional check to test if device supports discards? It would prevent test from failing until such mark would be added.

test/functional/tests/trim/test_trim.py
```diff
@@ -0,0 +1,105 @@
...
+
+    with TestRun.step("Stop blktrace and check if discard requests were issued"):
+        cache_reqs = blktrace.stop_monitoring()
+        cache_part_start = cas_part.begin
+
+        if len(cache_reqs) == 0:
+            TestRun.fail("No discard requests issued to the cas partition!")
+
+        non_meta_sector = (cache_part_start + metadata_size).get_value(Unit.Blocks512)
+        non_meta_size = cas_part.size - metadata_size
```
mmichal10, line 64: You are using `.get_value(Unit.Byte)` later anyway, why don't You call it here? It would be more consistent

test/functional/tests/trim/test_trim.py
```diff
@@ -0,0 +1,105 @@
...
+            if req.sector_number != non_meta_sector:
+                TestRun.fail(f"Discard request issued to wrong sector: {req.sector_number}, "
+                             f"expected: {non_meta_sector}")
+            if req.byte_count != non_meta_size.get_value(Unit.Byte):
+                TestRun.fail(f"Discard request issued with wrong bytes count: {req.byte_count}, "
+                             f"expected: {non_meta_size.get_value(Unit.Byte)}")
+
+        cas_fio.read_write(ReadWrite.read)
+        non_cas_fio.read_write(ReadWrite.read)
+        cas_fio.verification_with_pattern("0x00")
```
mmichal10, line 75: How about adding verification for second partition?

test/functional/tests/trim/test_trim.py
micrakow: Discard was triggered by CAS itself during cache start

mmichal10: Could You please add a comment with this information?

test/functional/tests/conftest.py
micrakow: I'll make it part of update-test-framework commit

test/functional/tests/trim/test_trim.py
micrakow: It's being done at line :82 (after setting io dir to read at line :74)

test/functional/tests/trim/test_trim.py
micrakow: Actually this todo can be removed since here cache will be optane or nand disk and it always supports trim (please correct me if I'm wrong)

test/functional/tests/trim/test_trim.py
micrakow: This would require changes in BlkTrace module to allow it to monitor more than 1 device at once which is currently not possible :(. I would leave it as it is for now and I'll try to add such funcionality (if I make it till end of june ;))

test/functional/tests/trim/test_trim.py
mmichal10: It is kind of tricky ¯\_(ツ)_/¯

PR #449, opened 2020-06-22
Add test for devices' capabilities

katlapinka:

### DIFF
--- a/test/functional/tests/misc/test_device_capabilities.py
+++ b/test/functional/tests/misc/test_device_capabilities.py
@@ -1,0 +1,191 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import math
+import pytest
+import os
+from api.cas import casadm, cli_messages
+from core.test_run import TestRun
+from storage_devices.disk import DiskTypeSet, DiskType, DiskTypeLowerThan
+from storage_devices.partition import Partition
+from test_tools import disk_utils, fs_utils
+from test_utils.output import CmdException
+from test_utils.size import Size, Unit
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.require_plugin("scsi_debug")
+def test_device_capabilities():
+    """
+        title: Test whether CAS device capabilities are properly set.
+        description: |
+          Test if CAS device takes into consideration differences between devices which are used to
+          create it.
+        pass_criteria:
+          - CAS device starts successfully using differently configured devices.
+          - CAS device capabilities are as expected.
+    """
+
+    core_device = TestRun.disks['core']
+    max_io_size_path = os.path.join(disk_utils.get_sysfs_path(core_device.get_device_id()),
+                                    'queue/max_sectors_kb')
+    default_max_io_size = fs_utils.read_file(max_io_size_path)
+
+    iteration_settings = [{"device": "SCSI-debug module",
+                           "dev_size_mb": 1024, "logical_block_size": 512, "max_sectors_kb": 1024},
+                          {"device": "SCSI-debug module",
+                           "dev_size_mb": 1024, "logical_block_size": 512, "max_sectors_kb": 256},
+                          {"device": "SCSI-debug module",
+                           "dev_size_mb": 1024, "logical_block_size": 512, "max_sectors_kb": 128},
+                          {"device": "SCSI-debug module",
+                           "dev_size_mb": 2048, "logical_block_size": 2048, "max_sectors_kb": 1024},
+                          {"device": "standard core device",
+                           "max_sectors_kb": int(default_max_io_size)},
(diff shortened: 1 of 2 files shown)


test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+            TestRun.LOGGER.info(f"Cas device {capability} has proper value: {cas_val} "
+                                f"(cache: {cache_val}, core: {core_val})")
+    cache.stop()
+
+
+def set_max_io_size(device, new_max_io_size: Size):
+    TestRun.LOGGER.info(
+        f"Setting max_sectors_kb for device {device.get_device_id()} to {new_max_io_size}")
+    path = os.path.join(disk_utils.get_sysfs_path(device.get_device_id()), 'queue/max_sectors_kb')
+    fs_utils.write_file(path, str(int(new_max_io_size.get_value(Unit.KibiByte))))
```
mmichal10, line 220: How about moving it to TF?

test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+        with TestRun.step("Start cache and add prepared core device as core."):
+            cache, core, error_output = prepare_cas_device(cache_device, core_device)
+        with TestRun.step("Compare capabilities for CAS device, cache and core "
+                          "(or check proper error if logical sector mismatch occurs)."):
+            compare_capabilities(cache_device, core_device, cache, core, error_output)
+        with TestRun.step("Recreate CAS device with switched cache and core devices."):
+            cache, core, error_output = prepare_cas_device(core_device, cache_device)
+        with TestRun.step("Compare capabilities for CAS device, cache and core "
+                          "(or check proper error if logical sector mismatch occurs)."):
+            compare_capabilities(core_device, cache_device, cache, core, error_output)
```
mmichal10, line 36: How about adding empty lines between test groups? I think it would increase readability

test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,229 @@
...
+    scsi_debug_params = {
+        "delay": "0",
+        "virtual_gb": "200",
+        "dev_size_mb": str(dev_size_mb),
+        "sector_size": str(sector_size),
+        "physblk_exp": str(physblk_exp)
+    }
+    scsi_debug = TestRun.plugin_manager.get_plugin('scsi_debug')
+    scsi_debug.params = scsi_debug_params
+    scsi_debug.post_setup()
```
robertbaldyga, line 134: Running plugin `post_setup()` explicitly from the test doesn't seem right. If we need module reload functionality provided by plugin, then plugin should have dedicated method for that. It should not be done through hack on core plugin API.

test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,229 @@
...
+        core_dev_bs = disk_utils.get_block_size(core_device.get_device_id())
+        core = cache.add_core(core_device)
+        if cache_dev_bs > core_dev_bs:
+            TestRun.LOGGER.error(
+                f"CAS device started with cache device logical block size ({cache_dev_bs}) "
+                f"greater than core device logical block size ({core_dev_bs})")
+        return cache, core, None
+    except CmdException as e:
+        if cache_dev_bs <= core_dev_bs:
+            raise
```
mmichal10, line 151: If you rise exception, wouldn't if fail whole test run (I mean whole pytest instance) instead of failing just single test?

test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,229 @@
...
+        with TestRun.step("Start cache and add prepared core device as core."):
+            cache, core, error_output = prepare_cas_device(cache_device, core_device)
+        with TestRun.step("Compare capabilities for CAS device, cache and core "
+                          "(or check proper error if logical sector mismatch occurs)."):
+            compare_capabilities(cache_device, core_device, cache, core, error_output)
+        with TestRun.step("Recreate CAS device with switched cache and core devices."):
+            cache, core, error_output = prepare_cas_device(core_device, cache_device)
+        with TestRun.step("Compare capabilities for CAS device, cache and core "
+                          "(or check proper error if logical sector mismatch occurs)."):
+            compare_capabilities(core_device, cache_device, cache, core, error_output)
```
mmichal10, line 46: Each group vary only with input params - device size, bs and max_sectors - all steps are repeated each time. What do you think about putting redundant code into loop?

test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -0,0 +1,192 @@
...
+        if core_dev_sectors_num != core_sectors_num:
+            TestRun.LOGGER.error(
+                "Number of sectors in CAS device and attached core device is different.")
+            cache.stop()
+            return
+        cas_capabilities = measure_capabilities(core)
+        cache_dev_capabilities = measure_capabilities(cache_device)
+        core_dev_capabilities = measure_capabilities(core_device)
+
+        for _, (capability, method) in enumerate(capabilities.items()):
```
robertbaldyga, line 178: There is no point of using `enumerate()` if you ignore counter value anyway.

PR #464, opened 2020-07-15
Initialize tests cleanup

katlapinka:


test/functional/tests/initialize/test_negative_load.py
```diff
@@ -42,7 +41,7 @@ def test_load_occupied_id():
     cache1.stop()
 
     TestRun.LOGGER.info("Starting cache with default id on different device")
```
mmichal10, line 43: How about using TestRun.step()?

PR #471, opened 2020-07-21
Tests for flushing huge cache

Ostrokrzew:


test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,169 @@
...
+        if int(cache.get_statistics().usage_stats.dirty) < minimum_4KiB_blocks:
+            TestRun.fail("There are not enough dirty data in cache!")
+
+    with TestRun.step("Unmount core and stop cache with flush."):
+        core.unmount()
+        try:
+            # this operation could take few hours, depending on core disk
+            output = TestRun.executor.run(stop_cmd(str(cache.cache_id)), timedelta(hours=12))
+            if output.exit_code != 0:
+                raise CmdException("Failed to stop cache.", output)
```
katlapinka, line 96: I'd prefer TestRun.fail here, because exception can cause stopping the whole tests execution. And this is actually fail when casadm stop command returns non-zero status here

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,169 @@
...
+
+    with TestRun.step("Unmount core device."):
+        core_dev.unmount()
+        remove(mnt_point, True, True, True)
+
+
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_flush_over_400_gibibytes_without_fs(cache_mode):
```
katlapinka, line 117: maybe rename it to raw_device because it will be easier to find those tests which are running on raw device. Please update test title as well

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,169 @@
...
+        minimum_4KiB_blocks = int(Size(400, Unit.GiB).get_value(Unit.Blocks4096))
+        if int(cache.get_statistics().usage_stats.dirty) < minimum_4KiB_blocks:
+            TestRun.fail("There are not enough dirty data in cache!")
+
+    with TestRun.step("Stop cache with flush."):
+        try:
+            # this operation could take few hours, depending on core disk
+            output = TestRun.executor.run(stop_cmd(str(cache.cache_id)), timedelta(hours=12))
+            if output.exit_code != 0:
+                raise CmdException("Failed to stop cache.", output)
```
katlapinka, line 167: change to TestRun.fail

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,163 @@
...
+            .target(core)
+            .size(file_size)
+        )
+        fio.default_run_time = timedelta(hours=4)  # timeout for non-time-based fio
+        fio.run()
+
+    with TestRun.step("Check if dirty data exceeded 400 GiB."):
+        minimum_4KiB_blocks = int(Size(400, Unit.GiB).get_value(Unit.Blocks4096))
+        if int(cache.get_statistics().usage_stats.dirty) < minimum_4KiB_blocks:
+            TestRun.fail("There are not enough dirty data in cache!")
```
katlapinka, line 157: there is

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,163 @@
...
+
+    with TestRun.step("Write data to exported object."):
+        test_file_copy = test_file_main.copy(mnt_point + "test_file_copy")
+        test_file_copy.refresh_item()
+        sync()
+
+    with TestRun.step("Check if dirty data exceeded 400 GiB."):
+        minimum_4KiB_blocks = int(Size(400, Unit.GiB).get_value(Unit.Blocks4096))
+        if int(cache.get_statistics().usage_stats.dirty) < minimum_4KiB_blocks:
+            TestRun.fail("There are not enough dirty data in cache!")
```
katlapinka, line 88: there is

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,162 @@
...
+          Flush cache when amount of dirty data in cache with core with filesystem exceeds 400 GiB.
+        pass_criteria:
+          - Flushing completes successfully without any errors.
+    """
+    with TestRun.step("Prepare devices for cache and core."):
+        cache_dev = TestRun.disks['cache']
+        cache_dev.create_partitions([file_size * 1.05])
+        cache_part = cache_dev.partitions[0]
+        core_dev = TestRun.disks['core']
+        Udev.disable()
```
katlapinka, line 44: You have to check if cache device size is greater than file size and the same for core. As a core you can use raid volume.

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,162 @@
...
+from storage_devices.disk import DiskType, DiskTypeLowerThan, DiskTypeSet
+from test_tools.disk_utils import Filesystem
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import IoEngine, ReadWrite
+from test_tools.fs_utils import remove
+from test_utils.filesystem.file import File
+from test_utils.os_utils import sync, Udev
+from test_utils.size import Size, Unit
+
+file_size = Size(420, Unit.GiB)
```
katlapinka, line 22: I assume that in original test filesize was ~600GB, because some day there was a problem with such size, so I'd like you to increase this value in this test as well.

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,162 @@
...
+
+    with TestRun.step("Unmount core device."):
+        core_dev.unmount()
+        remove(mnt_point, True, True, True)
+
+
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_flush_over_400_gibibytes_raw_device(cache_mode):
```
katlapinka, line 113: the same comments as in above test

katlapinka (changes requested):

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
Ostrokrzew: You're right, previously I probably looked at the description, which was out of date. I'll do what You ask for.

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
Ostrokrzew: I think that the device used as core (normal disk or raid) should be set in config, but checking the size is a good idea.

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,173 @@
...
+from test_tools.disk_utils import Filesystem
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import IoEngine, ReadWrite
+from test_tools.fs_utils import remove
+from test_utils.filesystem.file import File
+from test_utils.os_utils import sync, Udev
+from test_utils.size import Size, Unit
+
+file_size = Size(640, Unit.GiB)
+disk_size = file_size * 1.02
```
katlapinka, line 24: I'd rather call it something like "required_disk_size"

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,173 @@
...
+bs = Size(64, Unit.MebiByte)
+mnt_point = "/mnt/cas/"
+
+
+@pytest.mark.parametrizex("fs", Filesystem)
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
```
katlapinka, line 30: shouldn't it be parametrizex?

test/functional/tests/lazy_writes/test_flush_huge_dirty_data.py
```diff
@@ -0,0 +1,173 @@
...
+    with TestRun.step("Delete test files."):
+        test_file_main.remove(True)
+        test_file_copy.remove(True)
+
+    with TestRun.step("Unmount core device."):
+        core_dev.unmount()
+        remove(mnt_point, True, True, True)
+
+
+@pytest.mark.parametrize("cache_mode", CacheMode.with_traits(CacheModeTrait.LazyWrites))
```
katlapinka, line 114: shouldn't it be parametrizex?

Deixx (approved):

katlapinka (approved):

PR #476, opened 2020-07-24
Test for handling discard request on huge CAS device

Ostrokrzew:

### DIFF
--- a/test/functional/tests/volumes/test_discard_on_huge_core.py
+++ b/test/functional/tests/volumes/test_discard_on_huge_core.py
@@ -1,0 +1,68 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import pytest
+
+from api.cas import casadm
+from storage_devices.disk import DiskType, DiskTypeSet
+from core.test_run import TestRun
+from test_utils.size import Size, Unit
+
+scsi_dev_size_gb = str(40 * 1024)
+
+
+@pytest.mark.os_dependent
+@pytest.mark.require_plugin("scsi_debug", virtual_gb=scsi_dev_size_gb, opts="1")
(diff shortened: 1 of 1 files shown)


test/functional/tests/volumes/test_discard_on_huge_core.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+
+from api.cas import casadm
+from api.cas.cache_config import CacheMode
+from storage_devices.disk import DiskType, DiskTypeSet
+from core.test_run import TestRun
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/cas"
+syslog_path = "/var/log/messages"
```
rafalste, line 15: You won't find this log file in Debian based distros, because they log everything in /var/log/syslog. To be covered in such scenario, you can add a simple check:
```
from test_tools.fs_utils import check_if_file_exists
syslog_path = "/var/log/messages"
if not check_if_file_exists(syslog_path):
    syslog_path = "/var/log/syslog"
```

But... I don't see this log file used anywhere in this test, so you may get rid of it and just keep the above in mind for the future. ;)

test/functional/tests/volumes/test_discard_on_huge_core.py
```diff
@@ -0,0 +1,70 @@
...
+from storage_devices.disk import DiskType, DiskTypeSet
+from core.test_run import TestRun
+from test_utils.size import Size, Unit
+
+scsi_dev_size_gb = str(40 * 1024)
+
+
+@pytest.mark.os_dependent
+@pytest.mark.require_plugin("scsi_debug", virtual_gb=scsi_dev_size_gb, opts="1")
+@pytest.mark.parametrizex("cache_mode", CacheMode)
```
katlapinka, line 19: I think there is no need to do this in more than one, default cache mode. It will make this test too long.

katlapinka (approved):

Deixx (approved):

PR #477, opened 2020-07-29
Fix issue of removing a core device whose core id is a subset of other core device's 

eriksun421:
Here is the exact issue this patch fixes. 

[root@localhost ~]# casadm -L
type    id   disk              status    write policy   device
cache   1    /dev/nvme10n1p1   Running   wt             -
├core   1    /dev/nvme1n1      Active    -              /dev/cas1-1
└core   10   /dev/nvme2n1      Active    -              /dev/cas1-10
[root@localhost ~]# 
[root@localhost ~]# cat /etc/mtab | grep cas
/dev/cas1-10 /mnt/disk1 xfs rw,seclabel,relatime,attr2,inode64,noquota 0 0
[root@localhost ~]# 
[root@localhost ~]# casadm -R -i 1 -j 1
Can't remove core 1 from cache 1. Device /dev/cas1-10 is mounted!

### DIFF
--- a/casadm/cas_lib.c
+++ b/casadm/cas_lib.c
@@ -1811,7 +1811,7 @@ int check_if_mounted(int cache_id, int core_id)
 	FILE *mtab;
 	struct mntent *mstruct;
 	char dev_buf[80];
-	int dev_buf_len;
+	int dev_buf_len, mnt_fsname_len;
 	if (0 <= core_id) {
 		/* verify if specific core is mounted */
 		snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-%d", cache_id, core_id);
@@ -1829,9 +1829,11 @@ int check_if_mounted(int cache_id, int core_id)
 	}
 
 	while ((mstruct = getmntent(mtab)) != NULL) {
+		mnt_fsname_len = strnlen(mstruct->mnt_fsname,
+				(sizeof("/dev/") + sizeof(mstruct->mnt_fsname)));
 		/* mstruct->mnt_fsname is /dev/... block device path, not a mountpoint */
 		if ((NULL != mstruct->mnt_fsname)
-		    && (strncmp(mstruct->mnt_fsname, dev_buf, dev_buf_len) == 0)) {
+		    && (strncmp(mstruct->mnt_fsname, dev_buf, max(dev_buf_len, mnt_fsname_len)) == 0)) {
 			if (core_id<0) {
 				cas_printf(LOG_ERR,
 					   "Can't stop cache instance %d. Device %s is mounted!\n",
--- a/casadm/ocf_env.h
+++ b/casadm/ocf_env.h
@@ -11,6 +11,7 @@
 #include "safeclib/safe_lib.h"
 
 #define min(x, y)  ({ x < y ? x : y; })
+#define max(x, y)  ({ x > y ? x : y; })
 
 #define ENV_BUG_ON(cond) ({ if (cond) exit(1); })
 


Ostrokrzew:
ok to test

casadm/cas_lib.c
```diff
@@ -1831,7 +1831,7 @@ int check_if_mounted(int cache_id, int core_id)
 	while ((mstruct = getmntent(mtab)) != NULL) {
 		/* mstruct->mnt_fsname is /dev/... block device path, not a mountpoint */
 		if ((NULL != mstruct->mnt_fsname)
-		    && (strncmp(mstruct->mnt_fsname, dev_buf, dev_buf_len) == 0)) {
+		    && (strcmp(mstruct->mnt_fsname, dev_buf) == 0)) {
```
Ostrokrzew, line 1834: I believe it is safer to leave `strncmp` here and add 1 to `def_buf_len` few lines above to include terminal character.
To be hypercorrect You could use there `env_strnlen` in place of actual `strnlen`.
I wrote 2 tests for case like Yours - https://github.com/Open-CAS/open-cas-linux/pull/479

robertbaldyga:
ok to test

eriksun421:
Hi @Ostrokrzew , 

I agree with you, "max(dev_buf_len, mnt_fsname_len)" is a good solution for this issue. 
I'll push an amended commit.

michalwy:
> `mnt_fsname_len = strnlen(mstruct->mnt_fsname, (sizeof("/dev/") + sizeof(mstruct->mnt_fsname)));` - that's all because `mstruct->mnt_fsname` shows as block device path, but `sizeof(mstruct->mnt_fsname)` doesn't count _/dev/_ prefix and we need full length so I added the `sizeof("/dev/")` to it.

sizeof(mstruct->mnt_fsname) would return size of pointer (char *) because mnt_fsname is defined as char pointer, not as char array. Look how it is defined in mntent.h:

```
struct mntent
  {
    char *mnt_fsname;           /* Device or server for filesystem.  */
    char *mnt_dir;              /* Directory mounted on.  */
    char *mnt_type;             /* Type of filesystem: ufs, nfs, etc.  */
    char *mnt_opts;             /* Comma-separated options for fs.  */
    int mnt_freq;               /* Dump frequency (in days).  */
    int mnt_passno;             /* Pass number for `fsck'.  */
  };
```

So the calculation based on two sizeofs would always return the same value, no matter what actual string contains: 8 for pointer size + size of '/dev/' string. This rather not what you want to get here.

casadm/cas_lib.c
```diff
@@ -1829,9 +1829,11 @@ int check_if_mounted(int cache_id, int core_id)
 	}
 
 	while ((mstruct = getmntent(mtab)) != NULL) {
+		mnt_fsname_len = strnlen(mstruct->mnt_fsname,
+				(sizeof("/dev/") + sizeof(mstruct->mnt_fsname)));
```
michalwy, line 1833: This would always return the same value which is not buffer length. See my comment with explanation above.

Ostrokrzew: Do You think that `(sizeof("/dev/") + sizeof(mstruct->mnt_fsname))` could be replaced here with `sizeof(dev_buf)`? CAS wouldn't compare longer strings anyway.

robertbaldyga: `sizeof(dev_buf)` is good enough. CAS device names should never be longer than this.

robertbaldyga (changes requested):

PR #488, opened 2020-08-05
Add trim stress test

katlapinka:
Needs vdbench API:
https://github.com/Open-CAS/test-framework/pull/186

### DIFF
--- a/test/functional/tests/io/trim/test_trim_stress.py
+++ b/test/functional/tests/io/trim/test_trim_stress.py
@@ -1,0 +1,88 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import os
+from datetime import timedelta
+import pytest
+from api.cas import casadm
+from api.cas.cache_config import CacheMode, CacheLineSize
+from core.test_run import TestRun
+from storage_devices.disk import DiskTypeSet, DiskType
+from test_tools.disk_utils import Filesystem
+from test_utils.size import Size, Unit
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand, DiskType.sata]))
+@pytest.mark.require_disk("core", DiskTypeSet([DiskType.optane, DiskType.nand, DiskType.sata]))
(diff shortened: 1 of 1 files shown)


test/functional/tests/io/trim/test_trim_stress.py
```diff
@@ -0,0 +1,85 @@
...
+        for d in core_devices:
+            cores.append(cache.add_core(d))
+
+    with TestRun.step("Create filesystem and mount CAS devices."):
+        directories = []
+        for counter, core in enumerate(cores):
+            core.create_filesystem(Filesystem(counter % len(Filesystem)))
+            mount_dir = os.path.join(mount_point, str(counter + 1))
+            directories.append(mount_dir)
+            core.mount(mount_dir, ["discard"])
```
Ostrokrzew, line 57: Method `mount` from `Core` (subclass of `Device`) takes 1 argument.

katlapinka: it automatically executes method from the parent class with 2 parameters (I've checked it with the debugger). I can add the second parameter to the mount method in Core class in other PR, but it is not necessary for this test to run properly

test/functional/tests/io/trim/test_trim_stress.py
```diff
@@ -0,0 +1,85 @@
...
+        cache.stop()
+
+
+def run_vdbench(directories):
+    vdbench = TestRun.plugin_manager.get_plugin('vdbench')
+    config = f"data_errors=1,validate=yes,messagescan=no,create_anchors=yes\n" \
+             f"fsd=default,depth=4,width=5,files=10,sizes=" \
+             f"(1k,10,2k,10,4k,10,8k,10,16k,10,32k,10,64k,10,128k,10,256k,10,512k,10)\n" \
+             f"fsd=fsd0,anchor={directories[0]}\nfsd=fsd1,anchor={directories[1]}\nfsd=fsd2," \
+             f"anchor={directories[2]}\nfsd=fsd3,anchor={directories[3]}\nfwd=fwd1,fsd=fsd*," \
```
Deixx, line 76: directories would be easy to add using enumerate(directory) - if someone changes number of cores, then hardcoded values will not handle that

PR #489, opened 2020-08-05
tests: Fix CAS version parsing

robertbaldyga:
As CAS version format changed to custom one, we need to parse it
in non-default way.

### DIFF
--- a/test/functional/api/cas/casadm_parser.py
+++ b/test/functional/api/cas/casadm_parser.py
@@ -12,10 +12,9 @@ from test_utils.size import parse_unit
 from storage_devices.device import Device
 from api.cas.cache_config import *
 from api.cas.casadm_params import *
+from api.cas.version import CasVersion
 from datetime import timedelta
 from typing import List
-
-from packaging import version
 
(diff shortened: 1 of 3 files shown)


test/functional/api/cas/casadm_parser.py
```diff
@@ -15,14 +15,17 @@
...
 from api.cas.cache_config import *
 from api.cas.casadm_params import *
 from test_utils.size import parse_unit
 
 
+class Version:
+    def __init__(self, version_string):
+        self.main, self.major, self.minor, self.pr, self.type = version_string.split('.')
+        self.base = f"{self.main}.{self.major}.{self.minor}"
+
```
katlapinka, line 28: Please look at the CasVersion class in /api/cas/version.py and consider merging it with your Version class

mmichal10:
Maybe we should consider using versioning compatible with `PEP 440`?

test/functional/api/cas/casadm_parser.py
robertbaldyga: Done.

PR #512, opened 2020-08-25
configure: add CAS_BLK_BIDI_RQ() macro

mmichal10:
Since commit 8b3238cabd50e27 in linux kernel removed blk_bidi_rq() marco, it
has to be wrapped in CAS `configure` script


robertbaldyga (approved):

robertbaldyga:
We decided not to include it in this release.

PR #513, opened 2020-08-26
Minor refactorings and updates

Ostrokrzew:


test/functional/tests/incremental_load/test_core_pool.py
```diff
@@ -77,12 +86,15 @@ def test_core_pool_exclusive_open():
         core_disk.create_partitions([Size(1, Unit.GibiByte)])
         core_dev = core_disk.partitions[0]
         core_dev.create_filesystem(Filesystem.ext4)
+
     with TestRun.step("Add core device to core device pool using --try-add flag."):
-        core = casadm.try_add(core_dev, 1)
+        core = casadm.try_add(core_dev, 1, 1)
```
katlapinka, line 91: here as well

katlapinka (changes requested):

test/functional/tests/incremental_load/test_core_pool.py
```diff
@@ -36,24 +37,32 @@ def test_attach_core_pool():
...
     with TestRun.step("Add core device."):
-        cache.add_core(core_dev)
+        core = cache.add_core(core_dev)
+
     with TestRun.step("Stop cache."):
         cache.stop()
+
     with TestRun.step("Add previously used core device to core pool using --try-add flag."):
-        first_core = casadm.try_add(core_dev, cache.cache_id)
+        first_core = casadm.try_add(core_dev, cache.cache_id, core.core_id)
```
katlapinka, line 51: I recently missed that - please left it as it was before, like I said in older comments

test/functional/tests/incremental_load/test_incremental_load.py
```diff
@@ -358,50 +412,66 @@ def test_print_statistics_inactive(cache_mode):
...
     with TestRun.step("Run IO."):
         run_fio([first_core.system_path, second_core.system_path])
+
     with TestRun.step("Print statistics and check if there is no inactive usage section."):
         active_stats = cache.get_statistics()
         check_if_inactive_section_exists(active_stats, False)
+
     with TestRun.step("Stop cache."):
-        cache.stop()
+        if cache.get_cache_mode() != CacheMode.WB:
```
Deixx, line 432: You should probably use cache mode traits (Lazy Writes)

test/functional/tests/incremental_load/test_incremental_load.py
```diff
@@ -358,50 +412,66 @@ def test_print_statistics_inactive(cache_mode):
...
+                                   cache.get_cache_mode() == CacheMode.PT)
         check_inactive_usage_stats(inactive_stats_before.inactive_usage_stats.inactive_clean,
                                    inactive_stats_after.inactive_usage_stats.inactive_clean,
                                    "inactive clean",
-                                   cache.get_cache_mode() in [CacheMode.PT, CacheMode.WB])
+                                   cache.get_cache_mode() == CacheMode.PT)
         check_inactive_usage_stats(inactive_stats_before.inactive_usage_stats.inactive_dirty,
                                    inactive_stats_after.inactive_usage_stats.inactive_dirty,
                                    "inactive dirty",
                                    cache.get_cache_mode() != CacheMode.WB)
```
Deixx, line 473: Can cache modes be replaced by cache mode traits in checks?

Deixx (changes requested):

test/functional/tests/incremental_load/test_incremental_load.py
```diff
@@ -358,50 +413,79 @@ def test_print_statistics_inactive(cache_mode):
...
+
     with TestRun.step("Check cache statistics section of inactive devices."):
         inactive_stats_after = cache.get_statistics()
         check_if_inactive_section_exists(inactive_stats_after)
         check_number_of_inactive_devices(inactive_stats_after, 1)
+        # criteria for checks below
+        insert_write_traits = CacheModeTrait.InsertWrite in cache_mode_traits  # WT, WB, WO
+        lazy_write_traits = CacheModeTrait.LazyWrites in cache_mode_traits  # WB, WO
+        lazy_writes_or_no_insert_write_traits = (not insert_write_traits
+                                                 or lazy_write_traits)  # WA, WB, WO
```
Deixx, line 474: Actually using traits has value for being universal, so if we add another cache mode, we won't have to go through all the tests and update conditions accordingly. Using traits and then commenting which cache modes are those, is setting us again in that "search through the code" spot - this time just to update the comments.
Already you've missed PT for 'lazy or no insert' condition ;)

Deixx (changes requested):

test/functional/tests/incremental_load/test_incremental_load.py
Ostrokrzew: You're right, these comments can easily be outdated. I'll get rid of them.

Deixx (approved):

katlapinka (approved):

PR #523, opened 2020-09-14
Automatic package generator improvements

rafalste:
1. Add DEB package creation.
2. Update RPM spec file:
    - add a few missing info fields
    - tune to work with SLES
3. A few minor improvements in the generator script.
4. Fix a typo in casctl manpage.

**Limitations:**
    - There is no way to pass kernel version to DEB package name within OS package building environment. Variable substitution is not supported in 'Package' control field, so kernel version info (for which modules were build against) is present in the 'Description' field.
TODO: package kernel modules with DKMS
    - Until RPM version 4.13 there is no way to implement logical OR in dependency fields in SPEC file. In order to preserve compatibility in creating RPMs on both SLES and RHEL based distributions, there has to be no build requirement on elfutils-libelf-devel or libelf-devel respectively. But since 'configure' script is checking if those headers are really needed at build time, this is no longer a problem.


utils/pckgen
```diff
@@ -13,19 +13,27 @@
...
 TEMP_TEMPLATE="opencas-${THIS}"
-DEPENDENCIES=(sed git mktemp rsync)
+DEPENDENCIES=(git mktemp rsync sed)
 # Dependencies for particular packages creation:
 DEPENDENCIES_TAR=(tar)
 DEPENDENCIES_ZIP=(zip)
-DEPENDENCIES_RPM=(tar rpmbuild)
+DEPENDENCIES_RPM=(rpmbuild tar)
 DEPENDENCIES_SRPM=("${DEPENDENCIES_RPM[@]}")
+DEPENDENCIES_DEB=(debuild dh fakeroot tar)
```
robertbaldyga, line 35: When `debuild` and `dh` are missing, the following message is printed:
```
--- Checking for dependencies
ERROR: pckgen: Dependencies not installed. You need to provide these programs first: debuild dh
```
which is only partially helpful, because `apt install debuild dh` fails:
```
E: Unable to locate package debuild
E: Unable to locate package dh
```
as those programs are provided by packages named `devscripts` and `debhelper` respectively.

robertbaldyga (approved):

mmichal10 (approved):

utils/pckgen
```diff
@@ -13,19 +13,27 @@
...
+CAS_LICENSE_NAME="BSD-3-Clause-Clear"
 SUPPORTED_FROM_VERSION="20.03"
 THIS=$(basename "$0")
 ARCH="$(uname -i)"
 SYS_INFO="/etc/os-release"
 SCRIPT_BASE_DIR=$(dirname $(realpath "$0"))
-BASE_SPEC="$SCRIPT_BASE_DIR/$CAS_NAME.spec.base"
+RPM_SPEC_FILE="$SCRIPT_BASE_DIR/$THIS.d/rpm/CAS_NAME.spec"
+DEB_CONTROL_FILES_DIR="$SCRIPT_BASE_DIR/$THIS.d/deb/debian"
+PACKAGE_MAINTAINER="Rafal Stefanowski <rafal.stefanowski@intel.com>"
```
robertbaldyga, line 26: Do you really want to receive all those emails from satisfied users? :)

PR #473, opened 2020-07-23
Negative test for loading CAS module with lack of memory

Ostrokrzew:
OS dependent test

PR #527, opened 2020-09-16
Test OpenCAS module memory consumption

Ostrokrzew:
Waits for #473


test/functional/tests/cli/test_cas_memory_usage.py
```diff
@@ -0,0 +1,136 @@
...
+
+    with TestRun.step("Measure memory usage with OpenCAS module."):
+        available_mem_with_cas = get_free_memory()
+        memory_used_by_cas = available_mem_before_cas - available_mem_with_cas
+        TestRun.LOGGER.info(
+            f"OpenCAS module uses {memory_used_by_cas.get_value(Unit.MiB):.2f} MiB of DRAM."
+        )
+
+    with TestRun.step("Validate amount of memory used by OpenCAS module"):
+        expected_memory_used = Size(8.0 * get_dut_cpu_number(), Unit.MiB)
```
katlapinka, line 57: Why are you using 8.0MiB in calculation, when there is 7.5MiB mentioned in test description? Please correct values.

test/functional/tests/cli/test_cas_memory_usage.py
```diff
@@ -16,7 +16,64 @@
...
+        description: |
+          Check that OpenCAS kernel module memory usage is in acceptable limits.
+        pass_criteria: |
+          - Loaded OpenCAS kernel module should not consume more than 110% of total memory
+          calculated by requirement formula:
+          7.5 MiB * active CPUs number
+    """
+    with TestRun.step("Disable caching and memory over-committing."):
+        disable_memory_affecting_functions()
+        drop_caches()
```
Deixx, line 36: Shouldn't this be called with DropCachesMode.ALL flag?

test/functional/tests/cli/test_cas_memory_usage.py
```diff
@@ -16,7 +16,64 @@
...
+            memory_used_by_cas.get_value() / expected_memory_used.get_value()
+        )
+        TestRun.LOGGER.info(f"Actual to expected usage ratio: {requirement_fulfillment_ratio:.2f}")
+        if requirement_fulfillment_ratio > 1.1:
+            if requirement_fulfillment_ratio < 1.4:
+                TestRun.LOGGER.warning("Memory usage corresponds to required limit "
+                                       "(between 110-140%% of expected value)")
+            else:
+                TestRun.LOGGER.error("Memory usage exceeded required limit "
+                                     "(over 140% of expected value)")
```
Deixx, line 72: This is not what the pass criterion states. Warning still counts as pass, so either this check needs to be changed, or the pass criterion.

test/functional/tests/cli/test_cas_memory_usage.py
```diff
@@ -16,7 +16,64 @@
...
+            f" {expected_memory_used.get_value(Unit.MiB):.2f} MiB of DRAM."
+        )
+        requirement_fulfillment_ratio = (
+            memory_used_by_cas.get_value() / expected_memory_used.get_value()
+        )
+        TestRun.LOGGER.info(f"Actual to expected usage ratio: {requirement_fulfillment_ratio:.2f}")
+        if requirement_fulfillment_ratio > 1.1:
+            if requirement_fulfillment_ratio < 1.4:
+                TestRun.LOGGER.warning("Memory usage corresponds to required limit "
+                                       "(between 110-140%% of expected value)")
```
Deixx, line 69: 140%% -> 140%

Deixx (changes requested):

test/functional/tests/cli/test_cas_memory_usage.py
Ostrokrzew: I modified the pass criterion, but it could be discussed because, since this test was primarily written, CAS has changed a lot (probably, maybe not so much, I don't know for sure). Maybe new requirements are required.

Deixx (commented):
Not sure if the formula for calculating memory consumption is still valid

katlapinka:
This test will be updated when there will be memory usage statistics for casadm implemented

PR #548, opened 2020-10-09
Introduce CAS kernel module parameters configuration

Deixx:


test/functional/api/cas/cache_config.py
```diff
@@ -263,6 +264,110 @@ def default_nhit_params():
...
+class UseIoScheduler(Enum):
+    off = 0
+    on = 1
+    DEFAULT = on
+
+
+class KernelParameters:
+    cut_off_DEFAULT = 1
+    queue_size_DEFAULT = 65536
+    unblock_size_DEFAULT = 60000
```
katlapinka, line 291: I think it will be better to make full names for those variables

Deixx: Done

katlapinka (approved):

test/functional/api/cas/casadm.py
```diff
@@ -23,7 +25,13 @@ def help(shortcut: bool = False):
 
 def start_cache(cache_dev: Device, cache_mode: CacheMode = None,
                 cache_line_size: CacheLineSize = None, cache_id: int = None,
-                force: bool = False, load: bool = False, shortcut: bool = False):
+                force: bool = False, load: bool = False, shortcut: bool = False,
+                kernel_params: KernelParameters = None):
+    if kernel_params is None:
+        kernel_params = KernelParameters()
```
Ostrokrzew, line 31: Why not pass new blank instance of `KernelParameters` instead of None by default? It would be initialized anyway.

Ostrokrzew (dismissed):

test/functional/api/cas/casadm.py
Deixx: Makes sense :D

Ostrokrzew (approved):

test/functional/api/cas/cache_config.py
```diff
@@ -263,6 +264,114 @@ def default_nhit_params():
...
+                self.max_writeback_queue_size_DEFAULT
+            )
+            and equal_or_default(
+                self.writeback_queue_unblock_size, other.writeback_queue_unblock_size,
+                self.writeback_queue_unblock_size_DEFAULT
+            )
+        )
+
+    @classmethod
+    def DEFAULT(cls):
```
robertbaldyga, line 335: Instead of having this method we can set default values for constructor arguments.

Deixx: Default values for constructor are None, because that reflects the default action when you load the kernel module with no parameters provided. We need DEFAULT() for reference, but we don't want to explicitly set parameters by default.

katlapinka (approved):

PR #559, opened 2020-10-29
Include headers needed for nvme structures

anatol:
It fixes following compilation issue:

  modules/cas_cache/utils/utils_nvme.c: In function ‘cas_nvme_get_nsid’:
  modules/cas_cache/utils/utils_nvme.c:27:28: error: ‘NVME_IOCTL_ID’ undeclared (first use in this function)
     27 |  ret = ioctl_by_bdev(bdev, NVME_IOCTL_ID, (unsigned long)NULL);
        |                            ^~~~~~~~~~~~~
  modules/cas_cache/utils/utils_nvme.c:27:28: note: each undeclared identifier is reported only once for each function it appears in
  modules/cas_cache/utils/utils_nvme.c: In function ‘cas_nvme_identify_ns’:
  modules/cas_cache/utils/utils_nvme.c:41:9: error: variable ‘cmd’ has initializer but incomplete type
     41 |  struct nvme_admin_cmd cmd = { };

### DIFF
--- a/modules/cas_cache/utils/utils_nvme.c
+++ b/modules/cas_cache/utils/utils_nvme.c
@@ -9,6 +9,7 @@
 
 #include <linux/ioctl.h>
 #include <linux/file.h>
+#include <linux/nvme_ioctl.h>
 
 
 int cas_nvme_get_nsid(struct block_device *bdev, unsigned int *nsid)


Ostrokrzew:
ok to test

modules/cas_cache/utils/utils_nvme.c
```diff
@@ -9,6 +9,7 @@
 
 #include <linux/ioctl.h>
 #include <linux/file.h>
+#include <linux/nvme_ioctl.h>
```
arutk, line 12: Looks like this header is not present in older kernels (e.g. 3.19).  We typically handle differences between kernels using the ./configure script. You might want to consider adding a config option (e.g. HAS_NVME_IOCTL_H) which would expand to 1 if nvme_ioctl.h is present and include the file here conditionally, based on #if HAS_NVME_IOCTL_H.

arutk (changes requested):

Deixx:
@CAS-Linux-Jenkins ok to test

Deixx:
@anatol, please add Signed-off-by: Name <email> to your commit, so our Jenkins will let you pass ;)

robertbaldyga:
This patch is no longer needed, as problem got solved in other way.

Issue #563, opened 2020-11-02
Removing a core device whose core id is a subset of other core device's fails


Ostrokrzew:
## Description

Like in title

## Expected Behavior

Core should be successfully removed

## Actual Behavior

Core is not removed from cache.

## Steps to Reproduce

1. Create cache with few cores
2. Cores IDs should start with the same digit, e.g. cas1-8 and cas1-830
3. Mount cores
4. Try to remove one core, e.g. 8

## Possible Fix

https://github.com/Open-CAS/open-cas-linux/pull/564

## Logs

Cannot remove the unmounted core.\nFailed to remove core.\nexit_code: 1\nstdout: \nstderr: Can't remove core 8 from cache 1. Device /dev/cas1-830 is mounted!

## Your Environment

* OpenCAS version (commit hash or tag): edc729bed8b9f6c44d767858b45dcaaec1532d3f
* Operating System: CentOS 7.6/ CentOS 7.7
* Kernel version: 3.10/ 5.4
* Cache device type (model): no matter
* Core device type (model): no matter
* Cache configuration:
* * Cache mode: (default: wt)
* * Cache line size: (default: 4)
* * Promotion policy: (default: always)
* * Cleaning policy: (default: alru)
* * Sequential cutoff policy: (default: full)
* Other (e.g. _lsblk_, _casadm -P_, _casadm -L_)

PR #564, opened 2020-11-02
Compare whole exported object path during core removal

Ostrokrzew:
Duplicates #477 because author of previous PR is unresponsive.
Closes https://github.com/Open-CAS/open-cas-linux/issues/563

### DIFF
--- a/casadm/cas_lib.c
+++ b/casadm/cas_lib.c
@@ -1807,15 +1807,14 @@ int check_if_mounted(int cache_id, int core_id)
 	FILE *mtab;
 	struct mntent *mstruct;
 	char dev_buf[80];
-	int dev_buf_len;
-	if (0 <= core_id) {
+	int difference = 0, error = 0;
+	if (core_id >= 0) {
 		/* verify if specific core is mounted */
 		snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-%d", cache_id, core_id);
 	} else {
 		/* verify if any core from given cache is mounted */
 		snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-", cache_id);
 	}
-	dev_buf_len = strnlen(dev_buf, sizeof(dev_buf));
 
 	mtab = setmntent("/etc/mtab", "r");
 	if (!mtab)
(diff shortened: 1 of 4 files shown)


casadm/cas_lib.c
```diff
@@ -1825,9 +1825,10 @@ int check_if_mounted(int cache_id, int core_id)
 	}
 
 	while ((mstruct = getmntent(mtab)) != NULL) {
+		mnt_fsname_len = strnlen(mstruct->mnt_fsname, sizeof(dev_buf));
 		/* mstruct->mnt_fsname is /dev/... block device path, not a mountpoint */
 		if ((NULL != mstruct->mnt_fsname)
-		    && (strncmp(mstruct->mnt_fsname, dev_buf, dev_buf_len) == 0)) {
+		    && (strncmp(mstruct->mnt_fsname, dev_buf, max(dev_buf_len, mnt_fsname_len)) == 0)) {
```
arutk, line 1831: this defeats the purpose of using strncmp here (although there are sufficient checks to make it safe anyways). For the sake of purity I would suggest strncpy_s(..., mnt_fsname_len, ..., dev_bug_len)

Ostrokrzew:
retest this please

casadm/cas_lib.c
```diff
@@ -1840,6 +1841,8 @@ int check_if_mounted(int cache_id, int core_id)
 			}
 			return FAILURE;
 		}
+		if (error)
+			return FAILURE;
```
robertbaldyga, line 1845: Don't we want to check `error` first? IMO we shouldn't assume that `difference` contains a meaningful value it `error` set to non-zero.

robertbaldyga (changes requested):

casadm/cas_lib.c
Ostrokrzew: good point

casadm/cas_lib.c
```diff
@@ -1825,9 +1825,12 @@ int check_if_mounted(int cache_id, int core_id)
 	}
 
 	while ((mstruct = getmntent(mtab)) != NULL) {
+		mnt_fsname_len = strnlen(mstruct->mnt_fsname, sizeof(dev_buf));
+		error = strncmp_s(mstruct->mnt_fsname, mnt_fsname_len, dev_buf, dev_buf_len, &difference);
```
arutk, line 1829: Couple comments:
1. strncm_s  requires max buffer sizes for both strings rather than exact "strnlen" of the string. 
2. mstruct->mnt_fsname is provided by standard library call, so it would be wise to assume that max buffer length is PATH_MAX. 
3. It looks that now this code would treat two string if equal if mnt_fsname is a prefix of /dev/cas*. This is not correct.

I would replace these two lines with simple:
error = strncmp_s(mstruct->mnt_fsname, PATH_MAX, dev_buf, dev_buf_len, &difference);

casadm/cas_lib.c
```diff
@@ -1825,9 +1825,11 @@ int check_if_mounted(int cache_id, int core_id)
 	}
 
 	while ((mstruct = getmntent(mtab)) != NULL) {
+		error = strncmp_s(mstruct->mnt_fsname, PATH_MAX, dev_buf, dev_buf_len, &difference);
```
arutk, line 1828: actually dev_buf_len should be replaced by sizeof(dev_buf)

adam-rutkowski (approved):

arutk (approved):

robertbaldyga (approved):

PR #565, opened 2020-11-02
Introduce DKMS support for DEB packages

robertbaldyga:

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -2,9 +2,9 @@
 # Copyright(c) 2012-2019 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
-# If KERNELRELEASE is defined, we've been invoked from the
+# If $(M) is defined, we've been invoked from the
 # kernel build system and can use its language.
-ifneq ($(KERNELRELEASE),)
+ifneq ($(M),)
 
 include $(M)/config.mk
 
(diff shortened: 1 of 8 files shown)


rafalste (approved):

imjfckm (approved):

modules/Makefile
```diff
@@ -16,8 +16,9 @@ obj-y += cas_disk/
 else
 
 OCFDIR=$(PWD)/../ocf
-KERNEL_DIR ?= "/lib/modules/$(shell uname -r)/build"
 PWD=$(shell pwd)
+KERNEL_VERSION ?= "$(shell uname -r)"
+KERNEL_DIR ?= "/lib/modules/$(KERNEL_VERSION)/build"
```
arutk, line 21: Looks like this breaks building CAS with kernel provided by source path (KERNEL_DIR=<kernel source> make ...). You can keep the old implementation and get the kernel version by "make -C $KERNEL_DIR kernelversion"

utils/pckgen.d/deb/debian/CAS_NAME-modules.dkms
```diff
@@ -0,0 +1,11 @@
+PACKAGE_NAME="<CAS_NAME>-modules"
```
arutk, line 1: How is <CAS_NAME> replaced with actual name? Will the file name be changed as well ("CAS_NAME-modules.dkms") ?

modules/Makefile
robertbaldyga: It doesn't. I intentionally left `?=` for both to make it possible to override either `KERNEL_VERSION` or `KERNEL_DIR`.

utils/pckgen.d/deb/debian/CAS_NAME-modules.dkms
robertbaldyga: Yep. This is done by `utils/pckgen` for all files in this directory.

modules/Makefile
robertbaldyga: Given that it's not used anywhere else it doesn't seem to be an issue.

arutk: It's confusing at least. If KERNEL_VERSION is not used anywhere then I don't see any point in this change

PR #568, opened 2020-11-09
Occupancy per ioclass

mmichal10:


test/functional/api/cas/ioclass_config.py
```diff
@@ -21,22 +21,25 @@
...
+DEFAULT_IO_CLASS_RULE = "unclassified"
 MAX_CLASSIFICATION_DELAY = timedelta(seconds=6)
 IO_CLASS_CONFIG_HEADER = "IO class id,IO class name,Eviction priority,Allocation"
 
 
 @functools.total_ordering
 class IoClass:
     def __init__(self, class_id: int, rule: str = '', priority: int = None,
-                 allocation: bool = True):
+                 allocation: str = "1.00"):
```
katlapinka, line 34: Why don't you use float instead of str?

test/functional/api/cas/ioclass_config.py
```diff
@@ -148,7 +151,7 @@ def create_ioclass_config(
         )
     if add_default_rule:
         output = TestRun.executor.run(
-            f'echo "0,unclassified,22,1" >> {ioclass_config_path}'
+            f'echo "0,unclassified,22,1.00" >> {ioclass_config_path}'
```
katlapinka, line 154: You can use DEFAULT variables in here so it could be easier to update when some of them will be changed in the future

test/functional/tests/io_class/test_io_class_eviction_priority.py
```diff
@@ -0,0 +1,173 @@
...
+from test_tools.disk_utils import Filesystem
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from test_utils.filesystem.file import File
+from test_utils.os_utils import drop_caches, DropCachesMode, sync, Udev
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("cache_mode", [CacheMode.WT])
```
katlapinka, line 30: I don't think it has to be passed as parametrize when there is only one value of this parameter

test/functional/tests/io_class/test_io_class_eviction_priority.py
```diff
@@ -0,0 +1,173 @@
...
+@pytest.mark.parametrize(
```
katlapinka, line 31: please use parametrizex

test/functional/tests/io_class/test_io_class_occupancy_remap.py
```diff
@@ -0,0 +1,157 @@
...
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from test_utils.filesystem.file import File
+from test_utils.os_utils import drop_caches, DropCachesMode, sync, Udev
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("cache_mode", [CacheMode.WB, CacheMode.WT])
+@pytest.mark.parametrize(
```
katlapinka, line 31: please use parametrizex

test/functional/tests/io_class/test_io_class_occupancy_remap.py
```diff
@@ -0,0 +1,157 @@
...
+        Dd()
+        .input("/dev/zero")
+        .output(f"{path}")
+        .count(num_ios)
+        .block_size(Size(1, Unit.Blocks4096))
+    )
+    TestRun.LOGGER.info(f"{dd}")
+    dd.run()
+    sync()
+    drop_caches(DropCachesMode.ALL)
```
katlapinka, line 149: those methods are the same for each test file. Consider adding some file with common methods used by those tests

test/functional/tests/io_class/test_io_class_occupancy_resize.py
```diff
@@ -0,0 +1,165 @@
...
+from test_utils.os_utils import drop_caches, DropCachesMode, sync, Udev
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("cache_mode", [CacheMode.WT])
+@pytest.mark.parametrize(
+    "cache_line_size", [CacheLineSize.LINE_4KiB, CacheLineSize.LINE_16KiB, CacheLineSize.LINE_64KiB]
+)
+@pytest.mark.parametrize("new_occupancy", [0, 20, 70, 100])
```
katlapinka, line 33: please use parametrizex and consider removing cache mode, since there is only one element in arguments list

test/functional/tests/io_class/test_io_class_occupancy_stress.py
```diff
@@ -0,0 +1,177 @@
...
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrize("cache_mode", [CacheMode.WT])
+@pytest.mark.parametrize(
+    "cache_line_size", [CacheLineSize.LINE_4KiB, CacheLineSize.LINE_16KiB, CacheLineSize.LINE_64KiB]
+)
+def test_ioclass_stress(cache_mode, cache_line_size):
+    """
+        title: Verify if cache doesn't loose ability to saturate fully
+        description:
+          Trigger IO to cache and verify if doesn't loose ability to saturate
```
katlapinka, line 38: if it doesn't

test/functional/tests/io_class/test_io_class_occupancy_stress.py
```diff
@@ -0,0 +1,177 @@
...
+                .set_unit(Unit.Blocks4096)
+            )
+
+            if not isclose(actuall_occupancy.value, occupancy_limit.value, rel_tol=0.1):
+                TestRun.LOGGER.error(
+                    f"Occupancy for ioclass {io_class.id} does not match. "
+                    f"Limit: {occupancy_limit}, actuall: {actuall_occupancy}"
+                )
+
+    for i in range(0, 6):
```
katlapinka, line 111: You can use "for _ in TestRun.iteration(range(0, 6))" so the log will be more readable

katlapinka (changes requested):

test/functional/api/cas/ioclass_config.py
mmichal10: Using string instead of float will allow us to create negative tests with invalid allocation values.

test/functional/tests/io_class/test_io_class_core_id.py
```diff
@@ -169,21 +169,21 @@ def prepare(filesystem, cores_number):
...
-        allocation=True,
+        allocation="1.00",
         rule="unclassified",
         ioclass_config_path=ioclass_config.default_config_file_path,
     )
     ioclass_config.add_ioclass(
         ioclass_id=1,
         eviction_priority=22,
-        allocation=True,
+        allocation="0.00",
```
katlapinka, line 179: shouldn't it be "1.00"?

test/functional/tests/io_class/test_io_class_eviction_priority.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+            part_to_evict_end_occupancy = get_io_class_occupancy(cache, io_class.id, percent=True)
+
+            # Since number of evicted cachelines is always >= 128, occupancy is checked
+            # with approximation
+            if not isclose(part_to_evict_end_occupancy, 0, abs_tol=4):
+                TestRun.LOGGER.error(
+                    f"Wrong percent of cache lines evicted from part {io_class.id}. "
+                    f"Meant to be evicted {io_class.max_occupancy*100}%, actaully evicted "
+                    f"{io_class.max_occupancy*100-part_to_evict_end_occupancy}%"
```
katlapinka, line 138: I think that there could be a check whether other partitions are not evicted

test/functional/tests/io_class/test_io_class_occupancy_load.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.parametrizex("cache_line_size", CacheLineSize)
+def test_ioclass_occuppancy_load(cache_line_size):
+    """
+        title: Load cache with occupancy limit specified
+        description: |
+          Start cache in WB, add ioclasses with a occupancy limit. Populate cache
```
katlapinka, line 29: I think that the last sentence will be satisfying as a description. Other things like starting cache in WB will be visible in test steps

test/functional/tests/io_class/test_io_class_occupancy_load.py
```diff
@@ -0,0 +1,181 @@
...
+          and stop it without a flush. Load cache and verify if occupancy limits
+          are loaded correctly and if each part has assigned apropriate number of
+          dirty blocks.
+        pass_criteria:
+          - Occupancy thresholds have correct values for each ioclass after load
+    """
+    with TestRun.step("Prepare disks"):
```
katlapinka, line 36: I'd rather call this step "Prepare CAS devices" or something like that, maybe "Start cache in Write-Back mode and add core device". Please do this in all tests.

katlapinka (changes requested):

test/functional/tests/io_class/test_io_class_core_id.py
mmichal10: Few lines earlier is following comment:
```
# To make test more precise all workload except of tested ioclass should be
# put in pass-through mode
```
I guess it was a bag introduced when the test was developed. I verified that is still passing with this change.

test/functional/tests/io_class/test_io_class_occupancy_load.py
mmichal10: I went with `Prepare CAS devices`

test/functional/tests/io_class/test_io_class_eviction_priority.py
mmichal10: Check added

katlapinka (approved):
Tests are now OK for me

katlapinka (dismissed):
Tests are now OK for me

casadm/cas_lib.c
```diff
@@ -2031,11 +2031,11 @@ void partition_list_line(FILE *out, struct kcas_io_class *cls, bool csv)
...
-	const char *allocation;
-	if (cls->info.cache_mode != ocf_cache_mode_pt)
-		allocation = csv ? "1" : "YES";
-	else
-		allocation = csv ? "0" : "NO";
+	float allocation;
+	char allocation_str[5];
+
+	allocation = ((float)cls->info.max_size)/100;
+	snprintf(allocation_str, sizeof(allocation_str), "%.2f", allocation);
```
robertbaldyga, line 2038: Imo that would be more adequate for fixed point number.
```
snprintf(allocation_str, sizeof(allocation_str), "%d.%d",
		cls->info.max_size/100, cls->info.max_size%100);
```

casadm/cas_lib.c
```diff
@@ -2157,6 +2161,74 @@ static inline const char *partition_get_csv_col(CSVFILE *csv, int col,
...
+	if (strnlen(allocation, MAX_STR_LEN) == 1) {
+		/* Case with '0' or '1' */
+		value = strtoul(allocation, NULL, 10);
+		if (value > 1)
+			return FAILURE;
+
+		*part_size = value * 100;
+
+		return SUCCESS;
+	} else if (strnlen(allocation, MAX_STR_LEN) == 4) {
```
robertbaldyga, line 2178: Maybe instead of making separate cases for all possible string lengths we could just parse this number as fixed point number with optional radix point and fractional part, so that we could support not only "1" or "1.00" but also "1.0" notion.

robertbaldyga (changes requested):

casadm/cas_lib.c
```diff
@@ -2031,11 +2031,11 @@ void partition_list_line(FILE *out, struct kcas_io_class *cls, bool csv)
 {
 	char buffer[128];
 	const char *prio;
-	const char *allocation;
-	if (cls->info.cache_mode != ocf_cache_mode_pt)
-		allocation = csv ? "1" : "YES";
-	else
-		allocation = csv ? "0" : "NO";
+	float allocation;
+	char allocation_str[5];
```
arutk, line 2035: should be 6 including terminating zero

casadm/cas_lib.c
arutk: simple sscanf_s(allocation, "%d.%d") would do the job, just make sure to check return value to know whether optional part after the dot was read

modules/cas_cache/classifier.c
```diff
@@ -534,7 +534,9 @@ static cas_cls_eval_t _cas_cls_extension_test(
 
 	/* First character of @extension is '.', which we don't want to compare */
 	len = strnlen(extension + 1, dentry->d_name.len);
```
arutk, line 536: len = strnlen(extension + 1, dentry->d_name.len - (extension  -entry->d_name.name));

arutk: even better: len = dentry->d_name.len - (extension  - entry->d_name.name) - 1;

arutk (changes requested):

test/functional/api/cas/ioclass_config.py
```diff
@@ -21,22 +21,25 @@
...
         self.rule = rule
         self.priority = priority
         self.allocation = allocation
 
     def __str__(self):
         return (f'{self.id},{self.rule},{"" if self.priority is None else self.priority}'
-                f',{int(self.allocation)}')
+                f',{self.allocation}')
 
     def __eq__(self, other):
```
arutk, line 44: not sure what is the purpose of __eq__, but maybe it would make sense to compare self.allocation and other.allocation as a number rather than string, so that "1" == "1.0". For your consideration

arutk (changes requested):

casadm/cas_lib.c
arutk: actually I don't see why not  "sscanf_s("allocation, "%f"); "

test/functional/api/cas/ioclass_config.py
mmichal10: In case of negative testing it may be valuable to have possibility to use a characters other than numeric

casadm/cas_lib.c
mmichal10: I used both `%s` and `%f` to make validation and parsing easier.

arutk (dismissed):

casadm/cas_lib.c
```diff
@@ -2157,6 +2160,34 @@ static inline const char *partition_get_csv_col(CSVFILE *csv, int col,
...
+			return FAILURE;
+
+	for (i = 0; i < strnlen(allocation, MAX_STR_LEN); i++) {
+		if (!isdigit(allocation[i]) && allocation[i] != '.')
+			return FAILURE;
+	}
+
+	res = sscanf(allocation, "%f", &alloc);
+	if (res != 1)
+		return FAILURE;
```
robertbaldyga, line 2180: `strtof()` can do all of this and even more. E.g. `buf=".1."` will not be catched by this construction as an error, while using `strtof()` will prevent those kind of errors.

robertbaldyga (changes requested):

robertbaldyga (approved):

arutk (approved):

PR #573, opened 2020-11-17
Adding test 'test_io_engines'

karolinavelkaja:


test/functional/tests/security/test_security_io_engines.py
```diff
@@ -0,0 +1,75 @@
+#
+# Copyright(c) 2019-2020 Intel Corporation
```
robertbaldyga, line 2: 2020

robertbaldyga (changes requested):

test/functional/tests/security/test_security_io_engines.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2019-2020 Intel Corporation
+# Copyright(c) 2020 Intel Corporation
```
robertbaldyga, line 2: It's a good practice to avoid having commits that fix problems introduced by other commits within the same pull request. Please squash those two commits into one.

robertbaldyga (changes requested):

test/functional/tests/security/test_security_io_engines.py
```diff
@@ -0,0 +1,75 @@
...
+
+        cache = casadm.start_cache(cache_dev, cache_mode, force=True)
+
+        TestRun.LOGGER.info(f"Create filesystem '{filesystem}' on '{core_dev.system_path}'")
+        core_dev.create_filesystem(filesystem)
+        core = cache.add_core(core_dev)
+        core.mount(mount_point)
+
+    with TestRun.step("Run 15 minutes FIO with data integrity check on CAS device \n"
+                      "using IO Engine from configuration."):
```
Ostrokrzew, line 52: Instead of "using IO Engine from configuration." maybe print "using IO Engine {io_engine}."?

test/functional/tests/security/test_security_io_engines.py
```diff
@@ -0,0 +1,75 @@
...
+        TestRun.LOGGER.info(f"Tested configuration: \n"
+                            f"cache mode: {cache_mode}, \n"
+                            f"file system: {filesystem} \n"
```
Ostrokrzew, line 55: It's nothing important, but space at the end of line is unnecessary.

test/functional/tests/security/test_security_io_engines.py
```diff
@@ -0,0 +1,75 @@
...
+               .write_percentage(30)
+               .verification_with_pattern()
+               .size(Size(1, Unit.GibiByte))
+               .cpus_allowed(get_dut_cpu_physical_cores())
+               .cpus_allowed_policy(CpusAllowedPolicy.split))
+        fio.run()
+
+    with TestRun.step("Stop cache."):
+        core.unmount()
+        cache.stop()
```
Ostrokrzew, line 75: Seems unnecessary.

test/functional/tests/security/test_security_io_engines.py
karolinavelkaja: Details are printed out below in message (including io engine), in test step I can't insert the variable

Ostrokrzew (changes requested):
I've got second thoughts about this test. There's no failure condition, so it will always pass or end with an exception. Please, put some check for errors somewhere.

test/functional/tests/security/test_security_io_engines.py
```diff
@@ -0,0 +1,71 @@
...
+               .run_time(timedelta(minutes=15))
+               .time_based()
+               .target(f"{mount_point}/fio_file")
+               .read_write(ReadWrite.randrw)
+               .write_percentage(30)
+               .verification_with_pattern()
+               .size(Size(1, Unit.GibiByte))
+               .cpus_allowed(get_dut_cpu_physical_cores())
+               .cpus_allowed_policy(CpusAllowedPolicy.split))
+        fio.run()
```
katlapinka, line 71: As Slawek mentioned, you should add some check so test could fail in case there is some data integrity issue. E.g. you can check the output of fio - fio is running with verification_with_pattern parameter (also you can add .do_verify()) - with those parameters it will end with exit code different from 0, when data integrity error will occur.

katlapinka (changes requested):

test/functional/tests/security/test_security_io_engines.py
karolinavelkaja: Parameter .verification_with_pattern() contains the '.do_verify()' check. Also the command fio.run() contains check of status code so (as discussed with Kasia already) this code doesn't need modifications.

karolinavelkaja:
> I've got second thoughts about this test. There's no failure condition, so it will always pass or end with an exception. Please, put some check for errors somewhere.

@Ostrokrzew verification is placed in fio.run() command, where exit code is checking :)

Ostrokrzew (approved):

katlapinka (approved):

katlapinka (changes requested):
Please move this test to "io" group and change the name to "test_io_engines"

Ostrokrzew (approved):

katlapinka (approved):

robertbaldyga (approved):

PR #577, opened 2020-11-18
Adding test 'test_fault_power_hit' and cli messages accordingly

karolinavelkaja:
Signed-off-by: Karolina Rogowska <karolina.rogowsky@gmail.com>

### DIFF
--- a/test/functional/api/cas/cli_messages.py
+++ b/test/functional/api/cas/cli_messages.py
@@ -23,6 +23,16 @@ start_cache_with_existing_metadata = [
     r"Error occurred, please see syslog \(/var/log/messages\) for details\."
 ]
 
+error_inserting_cache = [
+    r"Error inserting cache \d+"
+]
+
+reinitialize_with_force_or_recovery = [
+    r"Old metadata found on device\.",
+    r"Please load cache metadata using --load option or use --force to",
+    r" discard on-disk metadata and start fresh cache instance\."
+]
+
 remove_inactive_core = [
     r"Error while removing core device \d+ from cache instance \d+",
     r"Core device is in inactive state"
--- a/test/functional/tests/fault_injection/test_fault_power_hit.py
+++ b/test/functional/tests/fault_injection/test_fault_power_hit.py
@@ -1,0 +1,97 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import time
+
+import pytest
+
+from api.cas import casadm, cli, cli_messages
+from api.cas.cache_config import CacheMode
+from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from test_utils.output import CmdException
+from test_utils.size import Size, Unit
+
+log_path = "/var/log/opencas.log"
+wait_long_time = 180
+wait_short_time = 15
+
+
+@pytest.mark.parametrizex("cache_mode", CacheMode)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.require_plugin("power_control")
+def test_fault_power_hit(cache_mode):
+    """
+        title: Test with power hit.
+        description: |
+          Test if there will be no metadata initialization after wake up
+          - when starting cache without initialization.
+        pass_criteria:
+          - Start cache without re-initialization failed.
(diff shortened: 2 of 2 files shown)


test/functional/tests/fault_injection/test_fault_power_hit.py
```diff
@@ -0,0 +1,86 @@
...
+
+@pytest.mark.parametrizex("cache_mode", CacheMode)
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_fault_power_hit(cache_mode):
+    """
+        title: Test with  power hit.
+        description: |
+          Test if there will be no metadata initialization after wake up
+          - when starting cache without initialization.
```
mmichal10, line 30: Lines of description doesn't have to start with '-'

test/functional/tests/fault_injection/test_fault_power_hit.py
```diff
@@ -0,0 +1,86 @@
...
+        message_found = check_log(cli_messages.reinitialize_with_force_or_recovery)
+
+        if not message_found:
+            time.sleep(wait_time_180_s)
+            check_log(cli_messages.reinitialize_with_force_or_recovery)
+
+
+def check_log(expected_message):
+    """Read recent lines in log, look for given, expected message."""
+    cmd = f"tail -n 200 {log_path}"
```
Ostrokrzew, line 78: There may be some entries in the file. In case like that your method will result in a false pass. Try to find a way to read only new entries.

test/functional/api/cas/cli_messages.py
```diff
@@ -23,6 +23,19 @@
...
+    r"Old metadata found on device\.",
+    r"Please load cache metadata using --load option or use --force to",
+    r" discard on-disk metadata and start fresh cache instance\."
+]
+
+reinitialize_with_force_or_recovery = [
+    r"Old metadata found on device\.",
+    r"Please load cache metadata using --load option or use --force to",
+    r" discard on-disk metadata and start fresh cache instance\."
+]
```
Ostrokrzew, line 37: Should it be repeated? Isn't it better to split message and check for two separate? I think above message could be divided to even three messages and I bet one of them is passed to stdout, not stderr.

Ostrokrzew (changes requested):

test/functional/tests/fault_injection/test_fault_power_hit.py
karolinavelkaja: good point!

test/functional/api/cas/cli_messages.py
karolinavelkaja: I split messages as you suggested, good catch, thanks! I've double checked, they are both passed to stderr, test is passing;)

Ostrokrzew (approved):

test/functional/tests/fault_injection/test_fault_power_hit.py
```diff
@@ -0,0 +1,95 @@
...
+        cache_device_link = cache_dev.get_device_link("/dev/disk/by-id")
+
+        # mark last lines in log to check later only new entries
+        last_read_line = 1
+        log_lines = TestRun.executor.run_expect_success(
+            f"tail -qn +{last_read_line} {log_path}").stdout.splitlines()
+        last_read_line += len(log_lines)
+
+    with TestRun.step("Hard reset."):
+        power_control = TestRun.plugin_manager.get_plugin('power_control')
```
katlapinka, line 55: please add @pytest.mark with required power control plugin

katlapinka (changes requested):

Ostrokrzew (approved):

test/functional/tests/fault_injection/test_fault_power_hit.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            cache_dev.system_path = cache_device_link.get_target()
+            cache = casadm.load_cache(cache_dev)
+            TestRun.LOGGER.info(f"Cache device loaded correctly (as expected).")
+        except CmdException as e:
+            TestRun.LOGGER.error(f"Failed to load cache device. Exception: {e.output}")
+
+        time.sleep(wait_short_time)
+        message_found = check_log(last_read_line, cli_messages.reinitialize_with_force_or_recovery)
+
```
katlapinka, line 81: please add here some comment why the check is made twice

Ostrokrzew (changes requested):

Ostrokrzew (approved):

katlapinka (approved):

PR #581, opened 2020-11-25
fix for RAIDs cleanup in prepare

karolinavelkaja:
Signed-off-by: Karolina Rogowska <karolina.rogowsky@gmail.com>

### DIFF
--- a/test/functional/tests/conftest.py
+++ b/test/functional/tests/conftest.py
@@ -183,13 +183,16 @@ def base_prepare(item):
         raids = Raid.discover()
         for raid in raids:
             # stop only those RAIDs, which are comprised of test disks
-            if all(map(
-                    lambda d: d.system_path in [bd.system_path for bd in TestRun.dut.disks],
-                    raid.array_devices
-            )):
+            if all(map(lambda device:
+                       any(map(lambda disk_path:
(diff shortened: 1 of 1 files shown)


test/functional/tests/conftest.py
```diff
@@ -183,13 +183,16 @@ def base_prepare(item):
...
+            if all(map(lambda device:
+                       any(map(lambda disk_path:
+                               disk_path in device.system_path,
+                               [bd.system_path for bd in TestRun.dut.disks])),
+                       raid.array_devices)):
                 raid.umount_all_partitions()
                 raid.remove_partitions()
                 raid.stop()
+                for device in raid.array_devices:
+                    Mdadm.zero_superblock(device.system_path)
```
Ostrokrzew, line 195: If You put superblock erasing here, don't You want to remove it below (line 199)?

karolinavelkaja: both are important, we may clear disk twice that way but without both commands we could miss something to clear (e.g. if raid would be on partitions then we would clear just them, not drives, so second command take care of proper cleaning)

Ostrokrzew (approved):

Deixx (approved):

PR #599, opened 2020-12-17
adding to cache helper function for metadata max size

karolinavelkaja:
adding to api file 'memory.py' with helper functions for memory measurements

### DIFF
--- a/test/functional/api/cas/cache.py
+++ b/test/functional/api/cas/cache.py
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2019-2021 Intel Corporation
+# Copyright(c) 2019-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
@@ -55,6 +55,23 @@ class Cache:
             .lower()
         )
         return CacheStatus[status]
+
+    @staticmethod
+    def calculate_metadata_max_size(cache_size: Size, cache_line_size: CacheLineSize):
+        """Get memory size according to requirement """
+        max_number_of_cores = 4096
+        # constants here correspond with OCF source code, where the same formula is used;
+        # they are partially empirical data so there are no proper/characteristic names for them
+        constant = Size(100, Unit.MebiByte)
+
+        cache_line_count = cache_size / cache_line_size.value
+        per_cache_line_metadata = Size(68) + 2 * cache_line_size.value / max_number_of_cores
+        amount_of_dram = constant + cache_line_count * per_cache_line_metadata
+
+        # maximum value is 110% of calculated size
+        metadata_max_size = 1.10 * amount_of_dram
+
+        return metadata_max_size
 
     @property
     def size(self):


test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,20 @@ def get_status(self):
         status = self.get_statistics().config_stats.status.replace(' ', '_').lower()
         return CacheStatus[status]
 
+    def get_metadata_max_size(self, cache_line_size, number_of_cores=4096):
```
katlapinka, line 68: Maybe make this method static and instead of _self_ set _cache_size_ as the first parameter, so user would not have to have running cache to count metadata max size

katlapinka (approved):

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,20 @@ def get_status(self):
         status = self.get_statistics().config_stats.status.replace(' ', '_').lower()
         return CacheStatus[status]
 
+    @staticmethod
+    def get_metadata_max_size(cache_size, cache_line_size, number_of_cores=4096):
```
Ostrokrzew, line 69: I assume You expect `cache_size` and `cache_line_size` to be objects of type Size. Give us a hint.

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,20 @@ def get_status(self):
...
+        """Get memory size according to requirement """
+        constant = int(Size(100, Unit.MebiByte))
+
+        cache_line_count = int(cache_size.value) / int(cache_line_size.value)
+        per_cache_line_metadata = 68 + 2 * int(cache_line_size.value) / number_of_cores
```
Ostrokrzew, line 74: If `cache_size` and `cache_line_size` are Size's objects, then where do You check unit? These could be bytes or gibibytes. You should use the `get_value(Unit.MebiByte)` method rather than access their values directly (constant unit is MiB).

Ostrokrzew (approved):

Ostrokrzew (changes requested):
missclicked earlier

Ostrokrzew (approved):

rafalste (approved):

Ostrokrzew (approved):

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,22 @@ def get_status(self):
         status = self.get_statistics().config_stats.status.replace(' ', '_').lower()
         return CacheStatus[status]
 
+    @staticmethod
+    def get_metadata_max_size(cache_size: Size, cache_line_size: Size, number_of_cores=4096):
+        """Get memory size according to requirement """
+        constant = int(Size(100, Unit.MebiByte))
+        cache_size_value = int(cache_size.value)
```
katlapinka, line 72: I suggest using cache_size.get_value(Unit.MebiByte) to make sure that everything is counted with proper unit. Please use this method everywhere where you are getting the size's value

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,22 @@ def get_status(self):
...
+        cache_line_size_value = int(cache_line_size.value)
+
+        cache_line_count = cache_size_value / cache_line_size_value
+        per_cache_line_metadata = 68 + 2 * cache_line_size_value / number_of_cores
+        amount_of_dram = constant + cache_line_count * per_cache_line_metadata
+
+        # maximum value is 110% of calculated size
+        metadata_max_size = 1.10 * amount_of_dram
+
+        return Size(metadata_max_size)
```
katlapinka, line 82: Will the default unit be correct? Use the same Unit in Size constructor as in calculations

katlapinka (changes requested):

Deixx (dismissed):

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,22 @@ def get_status(self):
...
+        constant = int(Size(100, Unit.MebiByte))
+        cache_size_value = int(cache_size.get_value(Unit.MebiByte))
+        cache_line_size_value = int(cache_line_size.get_value(Unit.MebiByte))
+
+        cache_line_count = cache_size_value / cache_line_size_value
+        per_cache_line_metadata = 68 + 2 * cache_line_size_value / number_of_cores
+        amount_of_dram = constant + cache_line_count * per_cache_line_metadata
+
+        # maximum value is 110% of calculated size
+        metadata_max_size = 1.10 * amount_of_dram
```
rafalste, line 80: It would be great to assign all those hard coded numbers to some constant variables with descriptive names or at least add some comments for clarification, where those numbers came from.

robertbaldyga: Sadly they came from OCF source code, when we use exactly the same formula. We have no better names for those constants, this is partially empirical data.

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,22 @@ def get_status(self):
         status = self.get_statistics().config_stats.status.replace(' ', '_').lower()
         return CacheStatus[status]
 
+    @staticmethod
+    def get_metadata_max_size(cache_size: Size, cache_line_size: Size, number_of_cores=4096):
```
robertbaldyga, line 69: I'd rename it to `calculate_metadata_max_size()` as this function is not a getter at all. It produces result based only on input numbers.

test/functional/api/cas/cache.py
rafalste: Let's attach your response as a comment here then. ;)

test/functional/api/cas/cache.py
```diff
@@ -66,6 +65,23 @@ def get_status(self):
         status = self.get_statistics().config_stats.status.replace(' ', '_').lower()
         return CacheStatus[status]
 
+    @staticmethod
+    def calculate_metadata_max_size(cache_size: Size, cache_line_size: CacheLineSize,
+                                    number_of_cores=4096):
```
arutk, line 70: This is a constant, no point in making this a function parameter. I would rename it to "max_number_of_cores"

arutk (changes requested):

arutk:
retest this please

PR #600, opened 2020-12-17
Add support for kernels up to 5.8

robertbaldyga:


anatol:
Thank you @robertbaldyga for your help.

While we are on this topic, is there any chance opencas can be forward ported further, to 5.10 kernel? 5.10 has been released and it is going to be the next LTS kernel. It is the kernel that will end up in many distros.

robertbaldyga:
> Thank you @robertbaldyga for your help.
> 
> While we are on this topic, is there any chance opencas can be forward ported further, to 5.10 kernel? 5.10 has been released and it is going to be the next LTS kernel. It is the kernel that will end up in many distros.

@anatol For v20.12 we are short in time, but we will most likely do it for v21.03.

configure.d/1_vfs_ioctl.conf
```diff
@@ -0,0 +1,50 @@
...
+                unsigned long arg)
+        {
+            int error = -ENOTTY;
+
+            if (!file->f_op->unlocked_ioctl)
+                goto out;
+
+            error = file->f_op->unlocked_ioctl(file, cmd, arg);
+            if (error == -ENOIOCTLCMD)
+                error = -ENOTTY;
```
arutk, line 41: why ENOTTY?

modules/cas_disk/exp_obj.c
```diff
@@ -217,22 +217,54 @@ static int _casdsk_get_next_part_no(struct block_device *bd)
...
 	struct blkpg_ioctl_arg barg;
+	int result = 0;
+	int part_no;
+
+	bd_file = filp_open(dsk->path, 0, 0);
+	if (IS_ERR(bd_file))
+		return PTR_ERR(bd_file);
+
+	usr_bpart = cas_vm_mmap(NULL, 0, sizeof(bpart));
+	if (IS_ERR((void *)usr_bpart)) {
```
arutk, line 236: cas_vm_mmap does not return error pointer. you should check for negative value instead and not use IS_ERR/PTR_ERR here

arutk (changes requested):

modules/cas_disk/exp_obj.c
robertbaldyga: It does. Look at the example usages in the kernel code:
https://elixir.bootlin.com/linux/latest/source/arch/x86/kvm/x86.c#L10372
https://elixir.bootlin.com/linux/latest/source/fs/binfmt_flat.c#L569

configure.d/1_vfs_ioctl.conf
robertbaldyga: It's returned by `vfs_ioctl()` on kernels where this function is available, so I made it consistent.

configure.d/1_generic_acct.conf
```diff
@@ -30,68 +33,100 @@ apply() {
     case "$1" in
```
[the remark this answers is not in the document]
robertbaldyga: Done.

arutk (approved):

mmichal10 (approved):

anatol:
> @anatol For v20.12 we are short in time, but we will most likely do it for v21.03.

I filed a followup here https://github.com/Open-CAS/open-cas-linux/issues/605

PR #623, opened 2021-01-08
Fix few issues from security review

Ostrokrzew:
Change strnlen to sizeof in set_device_path()
Fix casadm's handling of listing caches for caches added outside casadm.
Remove unused `buff` from `print_slow_atomic_cache_start_info()`
Add comment about calling only after normalizing path

### DIFF
--- a/casadm/cas_lib.c
+++ b/casadm/cas_lib.c
@@ -540,11 +540,9 @@ const char *get_core_state_name(int core_state)
 /* check if device is atomic and print information about potential slow start */
 void print_slow_atomic_cache_start_info(const char *device_path)
 {
-	char buff[MAX_STR_LEN];
 	struct kcas_cache_check_device cmd_info;
 	int ret;
 
-	get_dev_path(device_path, buff, sizeof(buff));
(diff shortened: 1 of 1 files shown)


casadm/cas_lib.c
```diff
@@ -820,10 +820,19 @@ struct cache_device *get_cache_device(const struct kcas_cache_info *info, bool b
 	cache->expected_core_count = info->info.core_count;
 	cache->id = cache_id;
 	cache->state = info->info.state;
-	if (set_device_path(cache->device, sizeof(cache->device), info->cache_path_name,
-			    sizeof(info->cache_path_name)) != SUCCESS) {
-		free(cache);
-		return NULL;
+	if (by_id_path) {
+		if (set_device_path(cache->device, sizeof(cache->device), info->cache_path_name,
```
arutk, line 824: I suggest to simply copy info->cache_path_name. We expect it to be "by-id".

Ostrokrzew: Yeah, You're right, this works and simplifies function.

arutk:
taking over this change in PR 644

PR #650, opened 2021-01-21
LVM api (ocl)

karolinavelkaja:

### DIFF
--- a/test/functional/tests/example/example_test.py
+++ b/test/functional/tests/example/example_test.py
@@ -21,6 +21,7 @@ from test_utils.filesystem.file import File
 from test_utils.size import Size, Unit
 from test_utils.emergency_escape import EmergencyEscape
 from test_utils.fstab import add_mountpoint
+from storage_devices.lvm import Lvm
 
 
 def setup_module():
@@ -257,3 +258,39 @@ def test_example_emergency_escape():
     assert not escape.verify_trigger_in_log(
         dmesg_out
     ), "Emergency Escape trigger marker found in dmesg"
+
+
+@pytest.mark.require_disk("cache1", DiskTypeSet([DiskType.hdd, DiskType.hdd4k, DiskType.sata]))
+@pytest.mark.require_disk("cache2", DiskTypeSet([DiskType.hdd, DiskType.hdd4k, DiskType.sata]))
+def test_lvm_example():
+    """
+        title: Example test using LVM API.
+        description: Create and discover Logical Volumes.
+        pass_criteria:
+          - LVMs created.
+          - LVMs discovered.
+    """
+    with TestRun.step("Prepare"):
+        disk_1 = TestRun.disks['cache1']
+        disk_1.create_partitions([Size(8, Unit.GiB)] * 2)
+        test_disk_1 = disk_1.partitions[0]
+        test_disk_3 = disk_1.partitions[1]
+
+        disk_2 = TestRun.disks['cache2']
+        disk_2.create_partitions([Size(8, Unit.GiB)] * 2)
+        test_disk_2 = disk_2.partitions[0]
+        test_disk_4 = disk_2.partitions[1]
+
+    with TestRun.step("Create LVMs"):
+        lvm1 = Lvm.create(Size(5, Unit.GiB), [test_disk_1, test_disk_2], "lvm5")
+        lvm2 = Lvm.create(70, [test_disk_3, test_disk_4], "lvm7")
+
+    with TestRun.step("Discover LVMs"):
+        lvms = Lvm.discover()
+
+    with TestRun.step("Check if created LVMs were discovered"):
+        for lvm in (lvm1, lvm2):
+            if lvm not in lvms:
+                TestRun.LOGGER.error(f"Created LVM {lvm.volume_name} not discovered in system!")
+
+        TestRun.LOGGER.info(f"Created LVMs present in the system.")
--- a/test/functional/tests/conftest.py
+++ b/test/functional/tests/conftest.py
@@ -29,6 +29,7 @@ from test_tools.mdadm import Mdadm
 from test_tools.fs_utils import remove
 from log.logger import create_log, Log
 from test_utils.singleton import Singleton
+from storage_devices.lvm import Lvm, LvmConfiguration
 
 
 class Opencas(metaclass=Singleton):
@@ -243,6 +244,11 @@ def base_prepare(item):
         from storage_devices.drbd import Drbd
         if Drbd.is_installed():
             __drbd_cleanup()
+
+        lvms = Lvm.discover()
+        if lvms:
+            Lvm.remove_all()
+            LvmConfiguration.remove_filters_from_config()
 
         raids = Raid.discover()
         for raid in raids:


karolinavelkaja:
PR needs: https://github.com/Open-CAS/test-framework/pull/234

Ostrokrzew (approved):

test/functional/tests/example/example_test.py
```diff
@@ -116,3 +117,39 @@ def test_create_example_files():
...
+        test_disk_3 = disk_1.partitions[1]
+
+        disk_2 = TestRun.disks['cache2']
+        disk_2.create_partitions([Size(8, Unit.GiB)] * 2)
+        test_disk_2 = disk_2.partitions[0]
+        test_disk_4 = disk_2.partitions[1]
+
+    with TestRun.step("Create LVMs"):
+        lvm = Lvm.create(Size(5, Unit.GiB), [test_disk_1, test_disk_2], "lvm5")
+        lvm2 = Lvm.create(70, [test_disk_3, test_disk_4], "lvm7")
```
rafalste, line 145: You could've use `lvm1` and `lvm2` here, so you wouldn't run out of names a few lines later. ;) But nevertheless, it works fine, so I'm good with it.

rafalste (approved):

test/functional/tests/example/example_test.py
karolinavelkaja: Correct, thanks! :D

test/functional/api/cas/casadm.py
```diff
@@ -61,6 +64,15 @@ def add_core(cache: Cache, core_dev: Device, core_id: int = None, shortcut: bool
     return core
 
 
+def add_lvm_as_core(cache: Cache, lvm: Lvm):
```
karolinavelkaja, line 67: As discussed with Kasia, this method will be removed and I will modify 'add_core' method to identify and use LVM as core.

test/functional/tests/example/example_test.py
```diff
@@ -116,3 +117,39 @@ def test_create_example_files():
...
+
+
+@pytest.mark.require_disk("cache1", DiskTypeSet([DiskType.hdd, DiskType.hdd4k, DiskType.sata]))
+@pytest.mark.require_disk("cache2", DiskTypeSet([DiskType.hdd, DiskType.hdd4k, DiskType.sata]))
+def test_lvm_example():
+    """
+        title: Example test using LVM API.
+        description: Create and discover Logical Volumes.
+        pass_criteria:
+          - LVM created.
```
Ostrokrzew, line 129: *LVMs

Ostrokrzew (changes requested):

Ostrokrzew (dismissed):

test/functional/api/cas/casadm.py
```diff
@@ -104,12 +106,16 @@ def stop_cache(cache_id: int, no_data_flush: bool = False, shortcut: bool = Fals
 
 def add_core(cache: Cache, core_dev: Device, core_id: int = None, shortcut: bool = False):
     _core_id = None if core_id is None else str(core_id)
+    if isinstance(core_dev, Lvm):
+        core_dev_path = resolve_to_by_id_link(core_dev.path)
```
arutk, line 110: I would prefer to require a lvm path to be provided by id instead.

arutk (approved):

arutk:
retest this please

test/functional/api/cas/casadm.py
karolinavelkaja: LVM class improved - now the "path" is the by-id path and we don't need changes for "add_core" function at all

test/functional/tests/common/lvm_common_utils.py
```diff
@@ -29,7 +30,8 @@ def run_fio_on_lvm(volumes: []):
 
 
 def get_test_configuration():
-    config = InitConfig.create_init_config_from_running_configuration()
+    InitConfig.create_init_config_from_running_configuration()
+    config_output = TestRun.executor.run(f"cat {opencas_conf_path}")
     devices = get_block_devices_list()
 
-    return config, devices
+    return config_output.stdout, devices
```
arutk, line 37: why is this better then the old version?

test/functional/tests/common/lvm_common_utils.py
```diff
@@ -1,13 +1,14 @@
 #
-# Copyright(c) 2021 Intel Corporation
+# Copyright(c) 2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
```
arutk, line 3: looks like you have two changes in one comit: introducing LVM and changes in get_test_configuration(). I suggest to split this into two commits.

test/functional/tests/common/lvm_common_utils.py
```diff
@@ -29,7 +30,8 @@ def run_fio_on_lvm(volumes: []):
```
arutk, line 30: the run_fio_on_lvm() function doesn't seem to do anything related to LVM. it seems to be more specific to some test scenarios (not integrated to the repo yet). I suggest removing it from here and moving to a test.

test/functional/tests/common/lvm_common_utils.py
arutk, line 37: Actually this isn't related to LVM either. Maybe it should be moved to init_config.py?

arutk (changes requested):

test/functional/tests/common/lvm_common_utils.py
karolinavelkaja: This function will be used by 3 tests  (https://github.com/Open-CAS/open-cas-linux/pull/697/files , https://github.com/Open-CAS/open-cas-linux/pull/682/files , https://github.com/Open-CAS/open-cas-linux/pull/694/files ) so placing it inside each tests would be code duplication and that's why it is implemented here.

test/functional/tests/common/lvm_common_utils.py
karolinavelkaja: Because comparing config on test level was no longer working - I suspect that config format changed: now we have inside cache config lines table and core config lines table. Parsing saved config file this way and using ".stdout" allows for simple comparison in test (comparison of configuration before and after reboot)

karolinavelkaja: That's true but it's used only by lvm related tests so looked like good "common place". I would say - if, in the future, someone would like to use it it can be moved of course:)

test/functional/tests/common/lvm_common_utils.py
robertbaldyga: TBH I don't mind code duplication if it makes the code easier to read. This helper function is not generic at all and it's just another function which I need to open in order to understand what exactly the test does. Please do not compress the code.

robertbaldyga (changes requested):

test/functional/tests/common/lvm_common_utils.py
karolinavelkaja: ok, I will change it :)

test/functional/api/cas/casadm.py
```diff
@@ -17,6 +17,8 @@
 from .casadm_params import *
 from .casctl import stop as casctl_stop
 from .cli import *
+from storage_devices.lvm import Lvm
+from test_utils.disk_finder import resolve_to_by_id_link
```
rafalste: Those two lines are added in the "lvm api (ocl)" commit and removed in the next one "lvm common: improve get_test_configuration function". No need to add them in the first place, especially that those imports are not used anywhere.

arutk (approved):

test/functional/tests/common/lvm_common_utils.py
karolinavelkaja: file lvm_common_utils.py will be removed and it's functions placed within corresponding tests:)

arutk (approved):

rafalste (approved):

Issue #665, opened 2021-01-29
For some float values casadm loads IO class config with wrong allocation value


katlapinka:
## Description
When I load IO class config file and then I list it, casadm shows wrong allocation values for some IO classes.
For example when I load IO class with allocation value set to 0.59, casadm says that allocation value for this IO class is equal to 0.58. The same happens also for 0.53 value, which after loading config bacomes 0.52

## Expected Behavior
Allocation value is the same in the config and in 'casadm -C --list' command output

## Actual Behavior
Allocation values are different in the config and in 'casadm -C --list' command output

## Steps to Reproduce
1. Create IO class config
2. Load IO class config
3. List IO class config using casadm

## Your Environment
* OpenCAS version (commit hash or tag): Open-CAS 20.12

PR #666, opened 2021-01-29
casadm: don't cast a float to an int explicitly

mmichal10:
If an underflowed float number was casted to an int explicitly, it's value was
rounded to a lower value.

Using `round()` function ensures that the number will be always rounded to a
nearest integer.


mmichal10:
Fixex #665

robertbaldyga (approved):

arutk (approved):

PR #678, opened 2021-02-03
doc: Introduce reqparse, the plain text requirement parser

robertbaldyga:

### DIFF
--- a/doc/reqparse
+++ b/doc/reqparse
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+#
+# Copyright(c) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import argparse
+import enum
+import glob
+import io
+import markdown
+import os
+import sys
+import yaml
+import yaml.scanner
+
+
+def error(filename, line, column, string):
+    string = "\n        ".join(string.split('\n'))
+    print(f"[ERROR] {filename}: line {line}, column {column}: {string}", file=sys.stderr)
+    exit(1)
+
+
+class Entity:
+    def __init__(self):
+        self.header_text = ""
+        self.header = None
+        self.text = ""
+
+
+class State(enum.Enum):
+    BEGIN = enum.auto()
+    GROUP_HEADER = enum.auto()
+    GROUP_TEXT = enum.auto()
+    REQ_HEADER_PRE = enum.auto()
+    REQ_HEADER = enum.auto()
+    REQ_TEXT = enum.auto()
+
+
+def parse_header(entity, filename, line):
+    try:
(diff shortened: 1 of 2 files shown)


doc/reqparse
```diff
@@ -0,0 +1,153 @@
...
+    if state == State.REQ_TEXT:
+            reqs.append(current_entity)
+    return (group, reqs)
+
+
+parser = argparse.ArgumentParser(prog="reqparse")
+parser.add_argument(
+    "-o", "--output", action="store",
+    help="Output file"
+)
```
Ostrokrzew, line 126: I know the user should explicitly specify output file, but what do You think about adding suffix dependent on output format?

Ostrokrzew (changes requested):

doc/reqparse
```diff
@@ -0,0 +1,153 @@
...
+    "-f", "--format", action="store",
+    choices=["html", "markdown"], default="markdown",
+    help="Output format {html|markdown}",
+)
+
+args = parser.parse_args(sys.argv[1:])
+
+buf = ""
+with io.StringIO() as output:
+    for filename in glob.glob("requirements/*"):
```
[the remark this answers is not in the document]
robertbaldyga: That was actually the plan.

doc/reqparse
robertbaldyga: This is not Linux-style behavior.

rafalste (approved):

Ostrokrzew (approved):

doc/reqparse
```diff
@@ -0,0 +1,153 @@
...
+        group, reqs = parse_file(filename)
+        output.write(f"# {group.header['group']}\n")
+        output.write(group.text)
+        for r in reqs:
+            output.write(f"## {r.header['title']}\n")
```
mmichal10, line 142: `###` instead of `##` would increase html's readability

doc/reqparse
```diff
@@ -0,0 +1,153 @@
...
+parser.add_argument(
+    "-f", "--format", action="store",
+    choices=["html", "markdown"], default="markdown",
+    help="Output format {html|markdown}",
+)
+
+args = parser.parse_args(sys.argv[1:])
+
+buf = ""
+with io.StringIO() as output:
```
mmichal10, line 136: How about adding a table of contents? It only requires initializing the `output` with `output.write("[TOC]\n")` and adding `toc` to list the list of extensions.

robertbaldyga: Let's rather make it with JS/CSS.

doc/reqparse
robertbaldyga: It's matter of CSS formatting.

PR #701, opened 2021-02-19
Add few management requirements

Ostrokrzew:
Add the requirements for the `casadm` commands (including the script commands):
* remove-core
* remove-inactive
* detach
* try-add


doc/requirements/management
```diff
@@ -0,0 +1,134 @@
+---
+group: Management
+---
+
+Management operations allows the user to control and maintain Open CAS.
+They are performed from console line with `casadm` management tool.
```
mmichal10, line 6: `casctl` can be used as well. I would avoid particular names because if the tools would be renamed/substituted or the management toolset would be extended with a additional tool, the document would require updating.

doc/requirements/management
```diff
@@ -0,0 +1,134 @@
...
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Remove active cache volume with dirty data with flush
+id: remove_cache_volume_active_flush
+---
+
+While cache volume is active and contains dirty data, the management tool shall
+allow to remove cache volume with flush.
```
mmichal10, line 15: I feel like the requirements are over specified. You added requirements for removing active&dirty and inactive core but you missed one for active&clean. 

I would start with the requirement for removing core (without mentioning about any particular state) because that's the feature that CAS shall be able to handle. After this general statement I would add the requirements for flushing a dirty core before its removal, for optional skipping that step by giving additional confirmation and for interrupting the flushing.

doc/requirements/management
```diff
@@ -0,0 +1,134 @@
...
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Disallow to interrupt removing active cache volume with dirty data
+without flush
+id: interrupt_removing_cache_volume_active_no_flush_disallow
+---
+
+When active cache volume that contains dirty data is removed without flush,
+the management tool shall not allow to interrupt the removal operation and the
+cache volume shall be successfully removed with proper warning about displayed.
```
mmichal10, line 47: If it is about management operations I would focus on writing requirements for what CAS shall support rather than what it shall not support. Btw with `When active cache volume that contains dirty data is removed without flush (...) the cache volume shall be successfully removed` you are implying that removing dirty core shall always succeed. I think it would be hard to implement

arutk:
comment regarding the title: "add *a* few management requirements"

PR #704, opened 2021-02-22
doc: Requirements for IO classes

mmichal10:


mmichal10:
retest this please

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+
+The product shall allow to configure up to 33 IO classes.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Attributes
+id: attributes
+---
+
+The IO class configuration shall include a following attributes for each IO class:
```
robertbaldyga, line 260: s/a following attributes/the following attributes/ (we use "the" for plural)

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+The product shall accept the eviction priorities in range 0-255 inclusive.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Empty eviction priority
+id: empty_priority
+---
+
+The product shall allow to omit eviction priority during specifying IO class
+configuration.
```
robertbaldyga, line 290: The product shall allow to omit eviction priority in IO class configuration file.

robertbaldyga, line 290: Btw it should be specified what happens when we omit this attribute. Does it take any default value? etc.

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Pinning IO class
+id: pinning
+---
+
+The product shall allow to pin IO classes.
```
robertbaldyga, line 298: What happens when the IO class is pinned?

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+Effective IO class for given IO request shall be the latest one for which
+the IO request fulfill the condition.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Last processed rule
+id: last_processed
+---
+
+The product shall allow to mark classification rule as the last eligible to process.
```
robertbaldyga, line 333: This is unclear what it means. It looks like when we mark rule as "last" then we stop processing rules unconditionally.

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Default IO class attribute
+id: default_ioclass_attribute
+---
+
+The product shall initialize default IO class with attributes:
+
+1. Id equal to 1
```
robertbaldyga, line 343: Isn't it 0?

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+2. Classification rule unclassified
```
robertbaldyga, line 344: It would be more readable if we surrounded "unclassified" with quotes.

robertbaldyga (changes requested):

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
 
-When initializing a cache the software shall set IO class's max occupancy to 1.00.
+When a target IO class max occupancy has been set to 0.00 the software shall submit
+IO requests in the Pass-Through mode.
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
-title: Setting an occupancy limit on loading a cache
-id: occupancy_limit_on_load
+title: Reassign cachelines to another IO class regardless of occupancy limit
```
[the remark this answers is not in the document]
arutk: never heard of "regargdless to". dictionary seems to confirm this: 
https://dictionary.cambridge.org/dictionary/english/regardless

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+title: Directory based classification
+id: directory_rule
 ---
 
-It shall be possible to display the currently set value of the occupancy limit
-for each configured ioclass. The value shall be displayed as a fraction of a cache
-size.
+The product shall allow to classify the file IO requests basing on an absolute
+path to a directory. The classification condition shall be fulfilled when the target
+file will be located in the given directory or in any level of its subdirectories.
```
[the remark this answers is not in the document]
arutk: agree, I don't think future tense would work anywhere in the requirements

doc/requirements/io_class
arutk: Agreed, I tnkink "pinning" is never used in user documentation

doc/requirements/io_class
arutk: Maybe something along these lines (not sure about the word "terminating"):
The product shall allow to mark a classification rule as a terminating one. Upon matching I/O to a terminating classification rule no further classification is performed and the resolved (resulting?) I/O class is the terminating one.

arutk (changes requested):

doc/requirements/io_class
```diff
@@ -19,65 +19,336 @@ It shall be possible to set a maximum size of an IO class.
...
+The product shall accept integer and fixed point values for a max size of an IO class
+in range from 0.00 to 1.00 and interpret them as a fraction of a cache size.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Rules testing order
+id: rules_testing_order
+---
+
+IO request shall be tested against classification rules in the order they were
```
arutk, line 315: Isn't this actually the order of increasing id?

arutk (changes requested):

doc/requirements/io_class
robertbaldyga: I think that we could point more explicitly that the termination takes place only if the IO request passes the check for the IO class marked as terminating. Btw this is a bit confusing that the "terminating" property of the IO class indicated by one of the classification rules.

doc/requirements/io_class
```diff
@@ -19,65 +19,350 @@ It shall be possible to set a maximum size of an IO class.
...
 
-When initializing a cache the software shall set IO class's max occupancy to 1.00.
+When a target IO class max occupancy has been set to 0.00 the software shall submit
+IO requests in the Pass-Through mode.
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
-title: Setting an occupancy limit on loading a cache
-id: occupancy_limit_on_load
+title: Reassign cachelines to another IO class regardless of occupancy limit
```
karolinavelkaja, line 41: probably "cache lines"?

doc/requirements/io_class
```diff
@@ -19,65 +19,350 @@ It shall be possible to set a maximum size of an IO class.
...
+The product shall provide an option to display current configuration of IO classes.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Export an IO class configuration to csv
+id: export_csv
+---
+
+The product shall provide an option to export current IO class configuration in
+the a format compatilbe with a config file.
```
karolinavelkaja, line 238: "in the a format" => without the

doc/requirements/io_class
```diff
@@ -19,65 +19,350 @@ It shall be possible to set a maximum size of an IO class.
...
+the IO request fulfill the condition.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Last processed rule
+id: last_processed
+---
+
+The product shall allow to mark classification rule as a terminating one. When
+an IO requests matches a terminating rule the IO class is resolved as an effecitve
```
karolinavelkaja, line 346: probably "an IO request" (singular)

robertbaldyga (approved):

doc/requirements/io_class
```diff
@@ -19,65 +19,348 @@ It shall be possible to set a maximum size of an IO class.
...
+The IO class with empty eviction priority shall be thought of as pinned.
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+title: Pinning IO class
+id: pinning
+---
+
+The product shall allow to pin an IO class. Data from the pinned IO class shall
+not be evicted unless its occupancy limit is not exceeded.
```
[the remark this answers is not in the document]
arutk: I think stating that something should not happen in an unspecified set of circumstances makes it hard to test. Instead I would specify a set of requirements exactly describing expected eviction/replacement behavior. This would also explain the interdependencies between specific settings (pinning, eviction priority, max occupancy) which might be hard to deduce from the existing form. Let's discuss this.

arutk (changes requested):

doc/requirements/io_class
```diff
@@ -19,65 +19,348 @@ It shall be possible to set a maximum size of an IO class.
...
 --------------------------------------------------------------------------------
-title: IO request to a disabled IO class
-id: occupancy_disabled_io_class
+title: Directory non existance detection
+id: rule_non_existing_directory
 ---
 
-When a target IO class max occupancy has been set to 0.00 the software shall submit
-IO requests in the Pass-Through mode.
+While a directory is not detected under the given path the product shall probe
```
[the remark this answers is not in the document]
mmichal10: You mean any delay introduced by CAS? Or by kernel?

doc/requirements/io_class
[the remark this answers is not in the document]
mmichal10: How exactly would you like to extend this req?

PR #707, opened 2021-02-23
Read hit tests

Ostrokrzew:


test/functional/tests/fault_injection/test_error_device_as_cache.py
```diff
@@ -0,0 +1,260 @@
...
+
+    with TestRun.step("Check if fio reported no read errors."):
+        if fio_errors != 0:
+            TestRun.fail("Fio reported read errors!")
+
+    with TestRun.step("Check cache error statistics."):
+        stats = cache.get_statistics()
+        read_errors_in_cache = stats.error_stats.cache.reads
+        total_errors_in_cache = stats.error_stats.cache.total
+        if read_errors_in_cache < fio_errors:
```
arutk, line 113: do you know why we check "<" instead of "!=" ?

Ostrokrzew: Actually, after some sleep, I think the checks in these two tests can be done in a different way. I will redesign some parts of both tests.

ArkadiuszNeumann:
retest this please

Deixx (commented):
commit minor fixes in tests can be dropped  
otherwise looks ok

PR #710, opened 2021-02-24
adding new lvm test: test_io_class_lvm_on_cas

karolinavelkaja:

### DIFF
--- a/test/functional/tests/io_class/test_io_class_lvm_on_cas.py
+++ b/test/functional/tests/io_class/test_io_class_lvm_on_cas.py
@@ -1,0 +1,130 @@
+#
+# Copyright(c) 2022 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+import pytest
+
+from api.cas import casadm, ioclass_config
+from api.cas.cache_config import CacheMode
+from api.cas.ioclass_config import IoClass
+from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from storage_devices.lvm import Lvm, LvmConfiguration
+from test_tools import fs_utils
+from test_tools.disk_utils import Filesystem
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/"
+io_target = "/mnt/test"
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_io_class_lvm_on_cas():
+    """
+        title: IO class for CAS device behind LVM.
+        description: Validate the ability of CAS to cache IO class when CAS device is used as LVM.
+        pass_criteria:
+          - Create CAS device and LVM on top of it successfully.
+          - Loading IO class configuration successfully.
+          - Running FIO for file size from IO class from 11 to 21 successfully.
+          - Increasing proper statistics as expected.
+    """
+
+    with TestRun.step(f"Create CAS device."):
+        cache_dev = TestRun.disks['cache']
+        core_dev = TestRun.disks['core']
+        cache_dev.create_partitions([Size(20, Unit.GibiByte)])
+        core_dev.create_partitions([Size(20, Unit.GibiByte)])
+
+        cache = casadm.start_cache(cache_dev.partitions[0], CacheMode.WB, force=True)
+        core = cache.add_core(core_dev.partitions[0])
+
+    with TestRun.step("Create LVM on CAS device."):
+        lvm_filters = ["a/.*/", "r|/dev/sd*|", "r|/dev/hd*|", "r|/dev/xvd*|", "r/disk/", "r/block/",
+                       "r|/dev/nvme*|"]
+
+        config = LvmConfiguration(lvm_filters,
+                                  pv_num=1,
+                                  vg_num=1,
+                                  lv_num=1,
+                                  cache_num=1,
+                                  cas_dev_num=1)
+
+        lvms = Lvm.create_specific_lvm_configuration(core, config)
+        lvm = lvms[0]
+
+    with TestRun.step("Create filesystem for LVM and mount it."):
+        lvm.create_filesystem(Filesystem.ext4)
+        lvm.mount(mount_point)
+
+    with TestRun.step("Prepare and load IO class config."):
+        io_classes = IoClass.csv_to_list(fs_utils.read_file("/etc/opencas/ioclass-config.csv"))
+        # remove two firs elements/lines: unclassified and metadata
+        io_classes.pop(1)
+        io_classes.pop(0)
+        IoClass.save_list_to_config_file(io_classes, add_default_rule=False)
+        cache.load_io_class(ioclass_config.default_config_file_path)
+
+    with TestRun.step("Run fio for file size from IO class from 11 to 21 "
+                      "and check that correct statistics increased."):
+        file_size = Size(2, Unit.KibiByte)
+
+        for io_class in io_classes:
+            if io_class.id < 11 or io_class.id > 21:
+                continue
+
+            TestRun.LOGGER.info(f"IO Class ID: {io_class.id}, class name: {io_class.rule}")
+            cache.reset_counters()
+
+            TestRun.LOGGER.info(f"Run FIO with verification on LVM [IO class ID {io_class.id}]")
+            (Fio().create_command()
+             .target(io_target)
+             .read_write(ReadWrite.randwrite)
+             .io_engine(IoEngine.libaio)
+             .io_depth(16)
+             .file_size(file_size)
+             .verification_with_pattern()
+             .write_percentage(100)
+             .block_size(Size(1, Unit.Blocks512))
+             .run())
+
+            TestRun.LOGGER.info(f"Checking statistics [IO class ID {io_class.id}]")
+
+            for io_class_i in io_classes:
+                class_stats = cache.get_io_class_statistics(io_class_i.id)
+                total_requests_io_class = class_stats.request_stats.requests_total
+
+                if io_class_i.id == io_class.id:
+                    if total_requests_io_class == 0:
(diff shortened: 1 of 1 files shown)


test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+        description: Validate the ability of CAS to cache IO class when CAS device is used as LVM.
+        pass_criteria:
+          - Create CAS device and LVM on top of it successfully.
+          - Loading IO class configuration successfully.
+          - Running FIO for file size from IO class from 11 to 21 successfully.
+          - Increasing proper statistics as expected.
+    """
+    with TestRun.step(f"Create CAS device."):
+
+        cache_dev, core_dev = prepare_devices()
```
Ostrokrzew, line 40: `prepare_devices()` returns lists, but the names of these variables are singular, not plural. Please change it to plural or return the first elements of both lists, you don't use more anyway. Or even close `cache` and `core` creation inside `prepare_devices()` and return them. IMO it would be more readable.

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+
+    with TestRun.step("Create filesystem for LVM and mount it."):
+        lvm.create_filesystem(Filesystem.ext4)
+        lvm.mount(mountpoint)
+
+    with TestRun.step("Load IO class config."):
+        cache.load_io_class("/etc/opencas/ioclass-config.csv")
+
+    with TestRun.step("Run fio for file size from IO class from 11 to 21 "
+                      "and check that correct statistics increase."):
```
Ostrokrzew, line 67: I suggest to split it to two steps to increase readability.

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+        csv = casadm.list_io_classes(cache.cache_id, OutputFormat.csv).stdout
+        io_classes = IoClass.csv_to_list(csv)
+        file_size = 2048
+        for io_class in io_classes:
+            if io_class.id < 11 or io_class.id > 21:
+                continue
+            print(f"IO Class ID: {io_class.id}, class name: {io_class.rule}")
```
Ostrokrzew, line 74: Use `TestRun.LOGGER.info` instead of `print`.

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+            cache.reset_counters()
+
+            print("Running FIO with verification on LVM.")
```
Ostrokrzew, line 77: ditto

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+            run_fio(file_size)
+
+            print("Checking statistics.")
```
Ostrokrzew, line 80: ditto

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+    return cache_device.partitions, core_device.partitions
+
+
+def run_fio(file_size):
+    (Fio().create_command()
+        .target(io_target)
+        .read_write(ReadWrite.randwrite)
+        .io_engine(IoEngine.libaio)
+        .io_depth(16)
+        .file_size(file_size)
```
Ostrokrzew, line 111: Fio's `file_size` method receives parameter of type `Size`. You use `int` in your code.

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+            print(f"IO Class ID: {io_class.id}, class name: {io_class.rule}")
+            cache.reset_counters()
+
+            print("Running FIO with verification on LVM.")
+            run_fio(file_size)
+
+            print("Checking statistics.")
+            check_statistics(cache, io_classes, io_class.id)
+
+            if file_size < 256 * 1024 * 1024:
```
Ostrokrzew, line 83: file_size < Size(256, Unit.MiB)

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
```diff
@@ -0,0 +1,151 @@
...
+                file_size *= 4
+            else:
+                file_size = 1100 * 1024 * 1024  # > 1 GiB
```
Ostrokrzew, line 86: file_size = Size(1, Unit.GiB)
Comment wouldn't be necessary.

Ostrokrzew (changes requested):

Deixx (changes requested):
Please rename test - DSS is a legacy name that no longer has meaning in CAS context. Generally it's "replaced" by IO class use.

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
karolinavelkaja: That way we can have both operations in one 'for' loop and checking statistics right after each FIO operation (before it we call cache.reset_counters for the statistics check). So running the workloads first and then checking statistics would impact test scenario/flow . Or did you mean something else by the split?

test/functional/tests/io_class/test_io_class_lvm_on_cas_dss.py
karolinavelkaja: it worked 🙈 but good catch, I will fix that, thanks!:)

karolinavelkaja:
This PR needs: https://github.com/Open-CAS/test-framework/pull/241

Ostrokrzew (approved):

Ostrokrzew (dismissed):

Ostrokrzew (approved):

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
```diff
@@ -0,0 +1,151 @@
...
+                TestRun.LOGGER.error(f"[{write_policy}] {stat_to_check} too low for IO Class "
+                                     f"{io_class.id} [{io_class.rule}]: {stat_value}")
+                TestRun.executor.run(f"ls -la {mountpoint}*")
+            else:
+                TestRun.LOGGER.info(f"[{write_policy}] {stat_to_check} for IO Class {io_class.id} "
+                                    f"[{io_class.rule}]: {stat_value}")
+                continue
+
+        if stat_value > 0:
+            TestRun.LOGGER.error(f"[{write_policy}] {stat_to_check} too high for IO Class "
```
rafalste, line 123: Maybe instead of "too low" and "too high", it will be better to explain exactly what went wrong - that statistics value is zero but should be increased or that it increase but should not.

rafalste:
retest this please

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def prepare_devices(cache_dev, core_dev, partitions_size: Size, cache_mode: CacheMode = None):
+    cache_dev.create_partitions([partitions_size])
+    core_dev.create_partitions([partitions_size])
+
+    cache = casadm.start_cache(cache_dev.partitions[0], cache_mode=cache_mode, force=True)
+    core = cache.add_core(core_dev.partitions[0])
+
+    return cache, core
```
robertbaldyga, line 140: This prepare is so simple, that you can put it in place (like you did in https://github.com/Open-CAS/open-cas-linux/pull/697).

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
```diff
@@ -0,0 +1,153 @@
...
+                                f"[IO class ID {io_class.id}].")
+            run_fio(file_size, io_target)
+
+            TestRun.LOGGER.info(f"Checking statistics [IO class ID {io_class.id}].")
+            check_statistics(cache, io_classes, io_class.id)
+
+            if file_size < Size(256, Unit.MebiByte):
+                file_size *= 4
+            else:
+                file_size = 1100 * Size(1, Unit.MebiByte)
```
robertbaldyga, line 88: Isn't it just `Size(1100, Unit.MebiByte)`?

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
```diff
@@ -0,0 +1,153 @@
...
+
+            fs_utils.remove(io_target)
```
robertbaldyga, line 90: Maybe I don't understand something. but I don't see how CAS device is actually used here.

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
```diff
@@ -0,0 +1,153 @@
...
+                                    f"[{io_class.rule}]: {stat_value}")
+                continue
+
+        if stat_value > 0:
+            TestRun.LOGGER.error(f"[{write_policy}] {stat_to_check} increased (not expected)"
+                                 f"for IO Class {io_class.id} [{io_class.rule}]: {stat_value}")
+            TestRun.executor.run(f"ls -la {mountpoint}*")
+        else:
+            TestRun.LOGGER.info(f"[{write_policy}] {stat_to_check} for IO Class {io_class.id} "
+                                f"[{io_class.rule}]: {stat_value}")
```
robertbaldyga, line 130: Why move it to separate function? It's called only once.

robertbaldyga (changes requested):

test/functional/tests/io_class/test_io_class_lvm_on_cas.py
karolinavelkaja: It was previously in "common" file. I've changed it, good point :)

robertbaldyga (approved):

KlaudiaJ (approved):

PR #712, opened 2021-02-25
Allocators revamp

imjfckm:
To improve allocations in OCF we need more flexible ways to allocate memory such as mpools. Also to reduce memory footprint we can reduce initial allocations for rpools backing the allocators.

### DIFF
--- a/modules/cas_cache/ocf_env.c
+++ b/modules/cas_cache/ocf_env.c
@@ -45,7 +45,9 @@ void *env_allocator_new(env_allocator *allocator)
 	struct _env_allocator_item *item = NULL;
 	int cpu;
 
-	item = cas_rpool_try_get(allocator->rpool, &cpu);
+	if (allocator->rpool)
+		item = cas_rpool_try_get(allocator->rpool, &cpu);
+
 	if (item) {
 		memset(item->data, 0, allocator->item_size -
 			sizeof(struct _env_allocator_item));
@@ -91,7 +93,8 @@ static void env_allocator_del_rpool(void *allocator_ctx, void *_item)
 
 #define ENV_ALLOCATOR_NAME_MAX 128
 
-env_allocator *env_allocator_create(uint32_t size, const char *name)
+env_allocator *env_allocator_create_extended(uint32_t size, const char *name,
+	int rpool_limit)
 {
 	int error = -1;
 	bool retry = true;
@@ -112,13 +115,6 @@ env_allocator *env_allocator_create(uint32_t size, const char *name)
 	}
 
 	allocator->item_size = size + sizeof(struct _env_allocator_item);
-	if (allocator->item_size > PAGE_SIZE) {
-		printk(KERN_WARNING "Creating allocator with item size"
-			" greater than 4096B");
-		ENV_WARN(true, OCF_PREFIX_SHORT" Creating allocator"
-			" with item size greater than 4096B\n");
-	}
-
 	allocator->name = kstrdup(name, ENV_MEM_NORMAL);
 
 	if (!allocator->name) {
@@ -156,13 +152,17 @@ RETRY:
 #endif
 
 	/* Initialize reserve pool handler per cpu */
+	if (rpool_limit < 0)
+		rpool_limit = CAS_ALLOC_ALLOCATOR_LIMIT;
 
-	allocator->rpool = cas_rpool_create(CAS_ALLOC_ALLOCATOR_LIMIT,
-			allocator->name, allocator->item_size, env_allocator_new_rpool,
-			env_allocator_del_rpool, allocator);
-	if (!allocator->rpool) {
-		error = __LINE__;
-		goto err;
+	if (rpool_limit > 0) {
+		allocator->rpool = cas_rpool_create(rpool_limit,
+				allocator->name, allocator->item_size, env_allocator_new_rpool,
+				env_allocator_del_rpool, allocator);
+		if (!allocator->rpool) {
+			error = __LINE__;
+			goto err;
+		}
 	}
 
 	return allocator;
@@ -172,6 +172,11 @@ err:
 	env_allocator_destroy(allocator);
 
 	return NULL;
+}
+
+env_allocator *env_allocator_create(uint32_t size, const char *name)
+{
+	return env_allocator_create_extended(size, name, -1);
 }
 
 void env_allocator_del(env_allocator *allocator, void *obj)
@@ -195,9 +200,11 @@ void env_allocator_del(env_allocator *allocator, void *obj)
 void env_allocator_destroy(env_allocator *allocator)
 {
 	if (allocator) {
-		cas_rpool_destroy(allocator->rpool, env_allocator_del_rpool,
-			allocator);
-		allocator->rpool = NULL;
+		if (allocator->rpool) {
+			cas_rpool_destroy(allocator->rpool, env_allocator_del_rpool,
+				allocator);
+			allocator->rpool = NULL;
+		}
 
 		if (atomic_read(&allocator->count)) {
 			printk(KERN_CRIT "Not all object deallocated\n");
--- a/modules/cas_cache/utils/utils_mpool.c
+++ b/modules/cas_cache/utils/utils_mpool.c
@@ -3,34 +3,60 @@
  * SPDX-License-Identifier: BSD-3-Clause-Clear
  */
 
-#include "ocf_env.h"
 #include "utils_mpool.h"
+#include "ocf_env.h"
 
+struct env_mpool {
+	int mpool_max;
+		/*!< Max mpool allocation order */
 
-struct cas_mpool *cas_mpool_create(uint32_t hdr_size, uint32_t size, int flags,
-		int mpool_max, const char *name_perfix)
+	env_allocator *allocator[env_mpool_max];
+		/*!< OS handle to memory pool */
+
+	uint32_t hdr_size;
+		/*!< Data header size (constant allocation part) */
+
+	uint32_t elem_size;
+		/*!< Per element size increment (variable allocation part) */
+
+	bool fallback;
+		/*!< Should mpool fallback to vmalloc */
+
+	int flags;
+		/*!< Allocation flags */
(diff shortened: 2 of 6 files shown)


modules/cas_cache/utils/utils_mpool.c
```diff
@@ -98,20 +124,24 @@ void *cas_mpool_new_f(struct cas_mpool *mpool, uint32_t count, int flags)
...
-	allocator = cas_mpool_get_allocator(mpool, count);
+	allocator = env_mpool_get_allocator(mpool, count);
 
 	if (allocator)
 		env_allocator_del(allocator, items);
-	else
+	else if (mpool->fallback)
 		cas_vfree(items);
+	else
+		return false;
```
arutk, line 144: ENV_BUG()? It seems that nobody checks the return value anyways, so you can change it back to void

modules/cas_cache/utils/utils_mpool.c
```diff
@@ -98,20 +124,24 @@ void *cas_mpool_new_f(struct cas_mpool *mpool, uint32_t count, int flags)
...
 }
 
-void cas_mpool_del(struct cas_mpool *mpool,
+bool env_mpool_del(struct env_mpool *mpool,
 		void *items, uint32_t count)
 {
 	env_allocator *allocator;
 
-	allocator = cas_mpool_get_allocator(mpool, count);
+	allocator = env_mpool_get_allocator(mpool, count);
```
arutk, line 137: early return in case items == NULL might be beneficial to avoid hitting the problematic case in line 144.

modules/cas_cache/utils/utils_mpool.h
```diff
@@ -6,55 +6,53 @@
...
+struct env_mpool;
 
-	env_allocator *allocator[cas_mpool_max];
-		/*!< OS handle to memory pool */
-
-	int flags;
-		/*!< Allocation flags */
+struct env_mpool_init_limit {
+	uint32_t size;
+	int rpool_limit;
```
arutk, line 30: maybe change the type to mpool item size enumeration (the one right above this struct)

modules/cas_cache/ocf_env.c
```diff
@@ -156,13 +152,17 @@ env_allocator *env_allocator_create(uint32_t size, const char *name)
...
 	/* Initialize reserve pool handler per cpu */
-
-	allocator->rpool = cas_rpool_create(CAS_ALLOC_ALLOCATOR_LIMIT,
-			allocator->name, allocator->item_size, env_allocator_new_rpool,
-			env_allocator_del_rpool, allocator);
-	if (!allocator->rpool) {
-		error = __LINE__;
-		goto err;
+	if (rpool_limit < 0)
+		rpool_limit = ENV_ALLOCATOR_NAME_MAX;
```
arutk, line 156: rpool_limit = ENV_ALLOC_ALLOCATOR_LIMIT ?

modules/cas_cache/utils/utils_mpool.c
```diff
@@ -3,34 +3,61 @@
...
-		mpool->allocator[i] = env_allocator_create(
-				hdr_size + (size * (1 << i)), name);
+		size = size_of(1 << i);
+
+		if (limits &&
+			limits[limit_iter].size &&
+			limits[limit_iter].size <= size) {
+			rpool_limit = limits[limit_iter].rpool_limit;
+			limit_iter++;
+		}
```
arutk, line 57: in "else" case you are stuck with rpool_limit  from previous loop iteration

modules/cas_cache/utils/utils_mpool.c
```diff
@@ -3,34 +3,61 @@
...
 
+	int flags;
+		/*!< Allocation flags */
+};
 
-struct cas_mpool *cas_mpool_create(uint32_t hdr_size, uint32_t size, int flags,
-		int mpool_max, const char *name_perfix)
+struct env_mpool *env_mpool_create(size_t (*size_of)(uint32_t), int flags,
+		int mpool_max, bool fallback,
+		const struct env_mpool_init_limit *limits,
```
arutk, line 25: The usage of "limits" is over complicated in my opinion. Does limits[i].size need to be one of "env_mpool_*"? If not, what is the purpose? If yes, let's just replace the limits with a simple array of integers indexed by "env_mpool_1":
     uint32_t limits[env_mpool_max]

arutk (changes requested):

modules/cas_cache/volume/vol_atomic_dev_bottom.c
```diff
@@ -1141,8 +1146,8 @@ int atomic_dev_init(void)
 	if (ret < 0)
 		return -EINVAL;
 
-	atomic_io_allocator = cas_mpool_create(0, sizeof(struct cas_atomic_io),
-			GFP_NOIO, 1, "cas_atomic_io");
+	atomic_io_allocator = env_mpool_create(atomic_io_size_of, GFP_NOIO, 1,
```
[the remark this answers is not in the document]
imjfckm: This will be used in OCF - `ocf_request` allocator has its own way of handling allocations exceeding max mpool order.

modules/cas_cache/utils/utils_mpool.c
imjfckm: OCF will use it while freeing requests, so it can determine whether it also has to free the map, but maybe I'll handle all the ocf_request-specific code in ocf_request instead of making mpools overly complicated.

modules/cas_cache/utils/utils_mpool.c
imjfckm: That was the intention - this way the limit will be applied to every allocation over the threshold until next threshold is exceeded

modules/cas_cache/utils/utils_mpool.c
imjfckm: My intention was to specify those limits in absolute size values, not in elements count. But maybe you're right that it's overly complicated for such a simple use case, so I'll probably change it

modules/cas_cache/utils/utils_mpool.c
imjfckm: I left it after all, let me know what you think about it.

modules/cas_cache/utils/utils_mpool.h
```diff
@@ -6,55 +6,53 @@
...
 /**
  * @brief Create CAS memory pool
  *
- * @param hdr_size Header size before array of items
- * @param size Size of particular item
+ * @param size_of getter for allocation size for given item count
  * @param flags Allocation flags
  * @param mpool_max Maximal allocator size (power of two)
+ * @param fallback Should allocations fall back to vmalloc if allocator fails
+ * @param limits Array of struct env_mpool_init_limit
```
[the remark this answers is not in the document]
imjfckm: Changed

modules/cas_cache/utils/utils_mpool.h
imjfckm: changed to simple array of limits

modules/cas_cache/ocf_env.c
imjfckm: Fixed

modules/cas_cache/utils/utils_mpool.c
imjfckm: Changed

modules/cas_cache/utils/utils_mpool.c
arutk: ok

arutk (approved):

arutk (dismissed):

modules/cas_cache/utils/utils_mpool.h
```diff
@@ -78,15 +74,17 @@ void *cas_mpool_new(struct cas_mpool *mpool, uint32_t count);
...
+void *env_mpool_new_f(struct env_mpool *mpool, uint32_t count, int flags);
 
 /**
  * @brief Free existing items of memory pool
  *
  * @param mpool CAS memory pool reference
  * @param items Items to be freed
  * @param count - Count of elements to be free
+ *
+ * @return Allocation was from mpool and was freed
```
rafalste, line 86: I think it returns true if allocation was freed regardless of mpool / vmalloc fallback.

modules/cas_cache/ocf_env.c
```diff
@@ -156,13 +152,17 @@ env_allocator *env_allocator_create(uint32_t size, const char *name)
...
-	allocator->rpool = cas_rpool_create(CAS_ALLOC_ALLOCATOR_LIMIT,
-			allocator->name, allocator->item_size, env_allocator_new_rpool,
-			env_allocator_del_rpool, allocator);
-	if (!allocator->rpool) {
-		error = __LINE__;
-		goto err;
+	if (rpool_limit < 0)
+		rpool_limit = CAS_ALLOC_ALLOCATOR_LIMIT;
+
+	if (rpool_limit > 0) {
```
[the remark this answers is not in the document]
imjfckm: Yup, the behavior here is:
* limit < 0 - allocate default rpool size
* limit == 0 - don't allocate rpools
* limit > 0 - allocate rpool in size different than defaults

modules/cas_cache/utils/utils_mpool.c
```diff
@@ -3,34 +3,60 @@
  * SPDX-License-Identifier: BSD-3-Clause-Clear
  */
 
-#include "ocf_env.h"
 #include "utils_mpool.h"
```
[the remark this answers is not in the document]
imjfckm: It's not necessary, but I think it's better to have explicit dependencies than implicit. Of course only to the point where you'd have to resolve circular dependencies where you don't need to ;)

modules/cas_cache/utils/utils_mpool.h
imjfckm: I removed the redundant info here to avoid confusion. Do you think it's better now?

rafalste (approved):

arutk (approved):

mmichal10 (approved):

PR #713, opened 2021-02-25
Update zero metadata API and tests

Ostrokrzew:

### DIFF
--- a/test/functional/api/cas/casadm.py
+++ b/test/functional/api/cas/casadm.py
@@ -153,9 +153,9 @@ def print_version(output_format: OutputFormat = None, shortcut: bool = False):
     return output
 
 
-def zero_metadata(cache_dev: Device, shortcut: bool = False):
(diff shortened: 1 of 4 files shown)


arutk:
These tests could be extended to cover all the cases considered in "zero_metadata" implementation:
1. zeroing metadata on a device containing dirty data (after clean shutdown) requires "--force"
2. zeroing metadatada after dirty shutdown requires "--force"
3. zeroing metadata works even if on-disk metadata is created in other version of CAS (not sure if it is feasible to implement such test in our framework)

Ostrokrzew:
I think you talking about functional tests for "zero-metadata" command. Tests, that are updated in this PR are not tests for this particular feature. Tests you're talking about should be designed and added to OCL.
1. and 2. point - easy to implement tests, we shall have a task for this;
3. - possible, but must wait for Michal's patch for CAS upgrade - there are many functionalities to simplify test-only upgrade.

arutk (approved):

karolinavelkaja (approved):

PR #736, opened 2021-03-16
fix warning in example_test.py

karolinavelkaja:

### DIFF
--- a/test/functional/tests/example/example_test.py
+++ b/test/functional/tests/example/example_test.py
@@ -16,7 +16,10 @@ from test_utils.size import Size, Unit
 
 
 def setup_module():
-    TestRun.LOGGER.warning("Entering setup method")
+    """
+    Function called by python
+    """
+    TestRun.LOGGER.info("Entering setup method")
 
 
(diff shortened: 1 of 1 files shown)


test/functional/tests/example/example_test.py
```diff
@@ -16,7 +16,10 @@
 
 
 def setup_module():
-    TestRun.LOGGER.warning("Entering setup method")
+    """
+    Function called by python
```
Ostrokrzew, line 20: *Python
This function is used nowhere anyway, so IMO can be dropped.

Ostrokrzew (changes requested):

test/functional/tests/example/example_test.py
karolinavelkaja: I thought so but during debug turned out that it is called by python.py

karolinavelkaja: so no - it should not be dropped and that's why the description had been added here.

Ostrokrzew (approved):

katlapinka:
retest this please

katlapinka (approved):

ArkadiuszNeumann:
retest this please

PR #747, opened 2021-03-22
Fix interruption tests

### LINKED ISSUE #749: tests need better flush detection
## Description
Tests: test_interrupt_cache_mode_switch_immediately and test_interrupt_cache_mode_switch_delayed need to be improved in the area of flush start detection

## Expected Behavior
flushing correctly detected by test and later checks performing in test are valid.

## Actual Behavior
Expecting flush operation to be started too early in test and later checks are invalid.

Ostrokrzew:
Add helper function and use it in interruption tests
Wait few seconds before reading cache statistics again
Update Test Framework

Fixes #749

### DIFF
--- a/test/functional/tests/fault_injection/test_fault_injection_interrupts.py
+++ b/test/functional/tests/fault_injection/test_fault_injection_interrupts.py
@@ -2,17 +2,21 @@
 # Copyright(c) 2020-2021 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
-from time import sleep
+
+from datetime import timedelta, datetime
 
 import pytest
 
 from api.cas import casadm, casadm_parser, cli
+from api.cas.cache import Cache
 from api.cas.cache_config import CacheMode, CleaningPolicy, CacheModeTrait
-from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from api.cas.core import Core
 from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
 from test_tools.disk_utils import Filesystem
 from test_utils import os_utils
 from test_utils.os_utils import Udev, DropCachesMode
+from test_utils.output import CmdException
 from test_utils.size import Size, Unit
 
 mount_point = "/mnt/cas"
@@ -34,7 +38,7 @@ def test_interrupt_core_flush(cache_mode, filesystem):
           - No system crash.
           - Flushing would be stopped after interruption.
           - Md5sum are correct during all test steps.
-          - Dirty blocks quantity after interruption is equal or lower.
+          - Dirty blocks quantity after interruption is equal or lower but non-zero.
     """
     with TestRun.step("Prepare cache and core."):
         cache_part, core_part = prepare()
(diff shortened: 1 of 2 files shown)


test/functional/tests/fault_injection/test_fault_injection_interrupts.py
```diff
@@ -70,12 +73,14 @@ def test_interrupt_core_flush(cache_mode, filesystem):
             sleep(2)
 
         with TestRun.step("Interrupt core flushing."):
+            wait_for_flushing(cache, core)
```
robertbaldyga, line 76: Now `sleep(2)` after running flush command is not needed.

test/functional/tests/fault_injection/test_fault_injection_interrupts.py
```diff
@@ -70,12 +73,14 @@ def test_interrupt_core_flush(cache_mode, filesystem):
...
             percentage = casadm_parser.get_flushing_progress(cache.cache_id, core.core_id)
             while percentage < 50:
                 percentage = casadm_parser.get_flushing_progress(cache.cache_id, core.core_id)
             TestRun.executor.run(f"kill -s SIGINT {flush_pid}")
 
         with TestRun.step("Check number of dirty data on exported object after interruption."):
+            sleep(4)
```
robertbaldyga, line 83: Why another sleep here?

test/functional/tests/fault_injection/test_fault_injection_interrupts.py
```diff
@@ -532,3 +546,15 @@ def create_test_file():
     dd.run()
     test_file.refresh_item()
     return test_file
+
+
+def wait_for_flushing(cache: Cache, core: Core, timeout=30):
```
robertbaldyga, line 551: Please use `timedelta`.

robertbaldyga (changes requested):

test/functional/tests/fault_injection/test_fault_injection_interrupts.py
```diff
@@ -532,3 +547,29 @@ def create_test_file():
...
+        except CmdException:
+            continue
+    raise CmdException("Management flush not started!")
+
+
+def wait_for_valid_dirty_blocks(volume, prev_stats,
+                                timeout: timedelta = timedelta(seconds=10)):
+    """
+    Necessary only in tests with the immediate interrupt because on some machines statistics read
+    by Test Framework are out-of-date and waiting few seconds solves the problem.
```
robertbaldyga, line 567: Do we know why is that?

Ostrokrzew: I have to admit I do not. I've set breakpoints and checked the statistics on DUT - there were changes, but in test I needed to wait up to 4 seconds. I ran test from PyCharm, but I'll try with command line, maybe IDE has some dummy cache.

ArkadiuszNeumann:
retest this please

karolinavelkaja (dismissed):

test/functional/tests/fault_injection/test_fault_injection_interrupts.py
karolinavelkaja: @robertbaldyga Please review again, it would be good to have this changes:)

robertbaldyga: My comment on that is still the same: I don't like this change. This is ugly workaround for a problem that exists somewhere else, be it test-framework or CAS itself. We expect stats to be updated immediately, so if they are not, then we'd rather like to have a test that fails than a test that introduces weird quirks only to obscure this problem.

Deixx: I've removed this method - during my tests on different platforms and OSes the problem described has not reappeared.

robertbaldyga (approved):

rafalste (approved):

PR #796, opened 2021-04-12
Remove inactive commands tests and API update

katlapinka:
Signed-off-by: klapinsk <katarzyna.lapinska@intel.com>

### DIFF
--- a/test/functional/api/cas/core.py
+++ b/test/functional/api/cas/core.py
@@ -36,7 +36,7 @@ class Core(Device):
         self.partitions = []
         self.block_size = None
 
-    def __get_core_info(self):
+    def __get_core_info(self, core_id=None):
         output = TestRun.executor.run(
             list_cmd(OutputFormat.csv.name, by_id_path=False))
         if output.exit_code != 0:
@@ -44,8 +44,9 @@ class Core(Device):
         output_lines = output.stdout.splitlines()
         for line in output_lines:
             split_line = line.split(',')
+            core_dev_id = self.core_device.get_device_id() if core_id is None else core_id
             if split_line[0] == "core" and (
-                    split_line[2] == os.path.join("/dev", self.core_device.get_device_id())
+                    split_line[2] == os.path.join("/dev", core_dev_id)
                     or split_line[5] == self.path):
                 return {"core_id": split_line[1],
                         "core_device": split_line[2],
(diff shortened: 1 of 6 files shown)


test/functional/tests/incremental_load/test_incremental_load.py
```diff
@@ -636,31 +637,53 @@ def test_remove_inactive_devices():
     with TestRun.step("Verify that all previously created CAS devices are listed with "
                       "proper status."):
         for core in cores:
-            if core.get_status() != CoreStatus.inactive:
+            if core.get_status(core_dev_id) != CoreStatus.inactive:
```
mmichal10, line 640: I wonder whether `.get_status(core_id)` should be a method of `Cache` class instead of `Core`. Each `Core` instance should be associated with one core device on DUT. With the current implementation you can check status of the core device which is not associated with the current `core` variable.

karolinavelkaja: "With the current implementation you can check status of the core device which is not associated with the current core variable." Correct but is that bad? I would say this is harmless and may be useful

Deixx: If the method is to be used like that, then it should be a class/static method. Current implementation is confusing and might lead to errors.

test/functional/api/cas/core.py
```diff
@@ -36,16 +36,17 @@ def __init__(self, core_device: str, cache_id: int):
...
         if output.exit_code != 0:
             raise Exception("Failed to execute list caches command.")
         output_lines = output.stdout.splitlines()
         for line in output_lines:
             split_line = line.split(',')
+            core_dev_id = self.core_device.get_device_id() if core_id is None else core_id
             if split_line[0] == "core" and (
-                    split_line[2] == os.path.join("/dev", self.core_device.get_device_id())
+                    split_line[2] == os.path.join("/dev", core_dev_id)
                     or split_line[5] == self.path):
```
Deixx, line 50: Mixing a parameter passed as an argument relating to a core and **self**, which also relates to a core (not necessarily the same) in one check is confusing.

Deixx (changes requested):

Deixx:
Closing this PR - continued in https://github.com/Open-CAS/open-cas-linux/pull/936

PR #802, opened 2021-04-15
Properly handle flush error in cache stop

### LINKED ISSUE #662: Flush error mishandling in cache stop
## Description
There are multiple problems with error handling in cache stop both in the driver as well as in casadm:
1. kernel driver would never return OCF_ERR_DIRTY_EXISTS because it is checking for dirty data after stopping the cache, at which point this information is lost.
2. casadm returns error exit status in case of KCAS_ERR_STOPPED_DIRTY, even when called with -n (no flush) parameter
3. when stopping the cache, runtime error during flush does not prevent the driver from stopping the cache anyway.

## Expected Behavior
1. stopping cache with -n (no flush) should return the following message if cache contained dirty data:
        {
                KCAS_ERR_STOPPED_DIRTY,
                "Cache has been stopped without fully flushing dirty data.\n"
                "Warning: Core devices may contain inconsistent data.\n"
                "To access your data, please start cache with --load option."
        },
2. successfully stopping cache with -n (no flush) should return exit code 0
3. I/O error during cache flush should result in cache remaining started.

## Actual Behavior
1. stopping cache with -n (no flush) does not print the expected message
2. I/O error during flush would not stop CAS from stopping the cache

## Steps to Reproduce
1. start cache in WB
2. add core
3. write some data (cache becomes diryt)
4. casadm -T -i 1 -n

for flush error handling verification some extra error injection would be needed

## Context
I suffer when I see this

## Possible Fix
Incomplete version of proposed fix available here:
https://github.com/Open-CAS/open-cas-linux/compare/65fe3152171600b0faf8039d3f068d7120d79baf...arutk:stop_err_hndl

## Your Environment

* OpenCAS version (commit hash or tag):
* Operating System:
* Kernel version:
* Cache device type (model):
* Core device type (model):
* Cache configuration:
* * Cache mode: (default: wt)
* * Cache line size: (default: 4)
* * Promotion policy: (default: always)
* * Cleaning policy: (default: alru)
* * Sequential cutoff policy: (default: full)
* Other (e.g. _lsblk_, _casadm -P_, _casadm -L_)

arutk:
In case of initial flush error stop is aborted. In case
of failure during the second flush, appropriate error
message is presented to the user.

Fixes #662 


arutk:
retest this please

arutk:
retest this please

mmichal10 (approved):

robertbaldyga (approved):

PR #811, opened 2021-04-29
Fix max discard size

mmichal10:
`blk_queue_max_discard_sectors()` accepts 32bit parameter


modules/cas_cache/volume/vol_block_dev_top.c
```diff
@@ -588,7 +588,8 @@ static void _blockdev_set_discard_properties(ocf_cache_t cache,
 		exp_q->limits.discard_granularity =
 			core_q->limits.discard_granularity;
 	} else {
-		blk_queue_max_discard_sectors(exp_q, core_sectors);
+		uint32_t max_discard_size = core_sectors <= UINT_MAX ?: UINT_MAX;
+		blk_queue_max_discard_sectors(exp_q, max_discard_size);
```
robertbaldyga, line 592: Will it be any problem if we always set it to `UINT_MAX` regardless of the device size?

modules/cas_cache/volume/vol_block_dev_top.c
```diff
@@ -588,7 +588,8 @@ static void _blockdev_set_discard_properties(ocf_cache_t cache,
 		exp_q->limits.discard_granularity =
 			core_q->limits.discard_granularity;
 	} else {
-		blk_queue_max_discard_sectors(exp_q, core_sectors);
+		uint32_t max_discard_size = core_sectors <= UINT_MAX ?: UINT_MAX;
```
arutk, line 591: what does it do? sets max_discard_size to UINT_MAX or 1?

mmichal10: I've meant `uint32_t max_discard_size = core_sectors <= UINT_MAX ? core_sectors : UINT_MAX`

modules/cas_cache/volume/vol_block_dev_top.c
mmichal10: `UINT_MAX` is set only if the device has more than `UINT_MAX` sectors.

modules/cas_cache/volume/vol_block_dev_top.c
arutk: I see. maybe use MIN macro?

mmichal10: done

PR #822, opened 2021-05-12
Fix adding core after core addition failure

### LINKED ISSUE #808: Sporadical problem with adding core device to the cache instance
## Description
There is a problem to add a core to the cache which results in error message like below:

> Error while adding core device to cache instance 1
> Cannot allocate memory

Problem occurs on RHEL 7.x

## Expected Behavior
Core should be added successfully

## Actual Behavior
This is impossible to add core

## Your Environment

* OpenCAS version (commit hash or tag):  21.03.0.0496.release
* Operating System: RHEL 7.9
* Kernel version: 3.10.0-1160.el7.x86_64
* Cache device type (NAND/Optane/other): Optane

Ostrokrzew:
caused by disallowed logical block size mismatch.

Cleanup exported object queue 
This patch fixes adding core after core addition failure.
The queue wasn't cleaned before and following core addition cannot
re-initialize queue properly.

Fixes #808

### DIFF
--- a/modules/cas_disk/exp_obj.c
+++ b/modules/cas_disk/exp_obj.c
@@ -816,7 +816,7 @@ int casdsk_exp_obj_destroy(struct casdsk_disk *dsk)
 }
 EXPORT_SYMBOL(casdsk_exp_obj_destroy);
 
-int casdsk_exp_obj_dettach(struct casdsk_disk *dsk)
+int casdsk_exp_obj_detach(struct casdsk_disk *dsk)
 {
 	module_put(dsk->exp_obj->owner);
 
--- a/modules/cas_disk/exp_obj.c
+++ b/modules/cas_disk/exp_obj.c
@@ -623,8 +623,7 @@ int casdsk_exp_obj_create(struct casdsk_disk *dsk, const char *dev_name,
 	return 0;
 
 error_set_geometry:
-	if (exp_obj->ops->cleanup_queue)
-		exp_obj->ops->cleanup_queue(dsk, queue, dsk->private);
+	blk_cleanup_queue(exp_obj->queue);
 error_init_queue:
 	blk_mq_free_tag_set(&dsk->tag_set);
 error_init_tag_set:
--- a/modules/cas_cache/cas_cache.h
+++ b/modules/cas_cache/cas_cache.h
@@ -70,7 +70,7 @@ extern ocf_ctx_t cas_ctx;
 extern struct casdsk_functions_mapper casdisk_functions;
 
 struct casdsk_functions_mapper {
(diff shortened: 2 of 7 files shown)


arutk (approved):

modules/cas_cache/volume/vol_block_dev_top.c
```diff
@@ -577,6 +577,18 @@ static void _blockdev_set_discard_properties(ocf_cache_t cache,
...
 }
 
+static void _blockdev_cleanup_queue(struct casdsk_disk *dsk)
+{
+	struct request_queue *exp_obj_q = NULL;
+
+	exp_obj_q = casdisk_functions.casdsk_exp_obj_get_queue(dsk);
+	BUG_ON(!exp_obj_q);
+
+	if (exp_obj_q) {
```
mmichal10, line 587: `if` is not necessary if you have `bug_on()`

modules/cas_disk/exp_obj.c
```diff
@@ -652,6 +651,17 @@ struct request_queue *casdsk_exp_obj_get_queue(struct casdsk_disk *dsk)
...
+void casdsk_exp_obj_cleanup_queue(struct casdsk_disk *dsk)
+{
+	struct request_queue *exp_obj_q = NULL;
+
+	exp_obj_q = casdsk_exp_obj_get_queue(dsk);
+	BUG_ON(!exp_obj_q);
+
+	blk_cleanup_queue(exp_obj_q);
+}
+EXPORT_SYMBOL(casdsk_exp_obj_cleanup_queue);
```
robertbaldyga, line 663: Why do we export this function? Is used only in this file, so it should be static.

robertbaldyga (changes requested):

modules/cas_disk/exp_obj.c
```diff
@@ -652,6 +651,17 @@ struct request_queue *casdsk_exp_obj_get_queue(struct casdsk_disk *dsk)
 }
 EXPORT_SYMBOL(casdsk_exp_obj_get_queue);
 
+void casdsk_exp_obj_cleanup_queue(struct casdsk_disk *dsk)
```
robertbaldyga, line 654: We can pass `queue` directly to this function instead of `dsk`. Btw this will make this function one-liner, so maybe just call `blk_cleanup_queue()` directly where it's used?

robertbaldyga (changes requested):

modules/cas_disk/exp_obj.c
Ostrokrzew: done

mmichal10 (approved):

PR #831, opened 2021-06-09
Set core's rotational field during core addition

Ostrokrzew:
Check if the backend device added to the cache is an HDD or not
during core addition so it would be done once.
Update OCF
Remove whitespace
Provide empty unused argument for management function


ArkadiuszNeumann:
retest this please

Deixx:
retest this please

PR #852, opened 2021-06-29
Fix fault injection tests

mmichal10:


test/functional/tests/fault_injection/test_soft_hot_plug_device.py
```diff
@@ -10,7 +10,7 @@
 import pytest
 
 from api.cas import casadm
-from api.cas.cache_config import CacheMode
```
robertbaldyga, line 13: It doesn't look like `CacheStatus` is used anywhere in this test.

test/functional/tests/fault_injection/test_soft_hot_plug_device.py
```diff
@@ -50,6 +50,8 @@ def test_soft_hot_plug_cache(cache_mode):
 
     with TestRun.step("Soft hot unplug cache device"):
         cache_dev.unplug()
+        if cache_dev.is_detected():
+            TestRun.block("Failed to unplug the device!")
```
robertbaldyga, line 54: This check should be performed by the TF itself.

test/functional/api/cas/statistics.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, stats):
             self.config_stats = CacheConfigStats(
                 *[stats[stat] for stat in config_stats_cache]
             )
-        except KeyError:
+        except KeyError as e:
```
robertbaldyga, line 55: This doesn't look necessary. `e` is not used.

robertbaldyga (changes requested):

test/functional/api/cas/statistics.py
Deixx, line 55: e not used

test/functional/tests/fault_injection/test_soft_hot_plug_device.py
```diff
@@ -10,7 +10,7 @@
 import pytest
 
 from api.cas import casadm
-from api.cas.cache_config import CacheMode
+from api.cas.cache_config import CacheMode, CacheStatus
```
Deixx, line 13: CacheStatus not used

Deixx (changes requested):

test/functional/tests/fault_injection/test_soft_hot_plug_device.py
mmichal10: Actually it is:
image
But if invalid serial has been set for a disk, TF reports the disk as already unplugged. After reconsidering the issue I will add a check to the function responsible for parsing configuration file.

PR #894, opened 2021-07-27
Async API for switching cleaning policy

mmichal10:


robertbaldyga:
Replaced with #896

PR #895, opened 2021-07-27
Fix lba classification test

### LINKED ISSUE #887: LBAs outside of a LBA range configured in IO class are cached
## Description

test_ioclass_lba 

## Expected Behavior

Only LBAs inside the configured range should be cached

## Actual Behavior

LBAs outside the configured range are cached

## Steps to Reproduce

1. Start cache and add core
2. Configure IO class
```
IO class id,IO class name,Eviction priority,Allocation
0,unclassified,255,0.00
1,lba:ge:56&lba:le:200&done,1,1.00
```
3. Write to LBAs inside and outside of configured LBA range
```dd if=/dev/zero of=/dev/cas1-1 count=1 bs=512 seek=72```

## Your Environment

* OpenCAS version (commit hash or tag): 21.06.0.0534.master b57a8bf19575cc3d2707f5953ad6f9ac9ab59fd9
* Operating System: Centos 7.9
* Kernel version: 5.10
* Cache device type (NAND/Optane/other): Optane
* Core device type (HDD/SSD/other): TLC NVMe

### PR DESCRIPTION
**Pull request by jfckm**
Fixes #887 

This test wasn't using `O_DIRECT` mode, so page cache was flushing the IO's with page granularity. LBA 200 lies on page start so if anything was written to up to LBA 207 then it would be improperly classified as pinned IO-class during pagecache flushing.

### DIFF
--- a/test/functional/tests/io_class/test_io_classification.py
+++ b/test/functional/tests/io_class/test_io_classification.py
@@ -39,7 +39,6 @@ def test_ioclass_lba():
     ioclass_id = 1
     min_cached_lba = 56
     max_cached_lba = 200
-    iterations = 100
     dd_size = Size(1, Unit.Blocks512)
     dd_count = 1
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/functional/tests/io_class/test_io_classification.py [hunk 1]
```diff
@@ -70,37 +70,40 @@ def test_ioclass_lba():
...
             dirty = cache.get_io_class_statistics(io_class_id=ioclass_id).usage_stats.dirty
             if dirty.get_value(Unit.Blocks4096) != dirty_count:
                 TestRun.LOGGER.error(f"LBA {lba} not cached")
 
-    with TestRun.step("Flush cache."):
-        cache.flush_cache()
-
     with TestRun.step("Run IO and check if lba outside of defined range are not cached."):
-        TestRun.LOGGER.info(f"Writing to random sectors outside of cached range.")
-        for i in range(iterations):
```
mmkayPL, line 87: iterations variable is unused now, remove it

test/functional/tests/io_class/test_io_classification.py
```diff
@@ -70,37 +70,40 @@ def test_ioclass_lba():
...
-            rand_lba = random.randrange(2000)
-            if min_cached_lba <= rand_lba <= max_cached_lba:
-                continue
+        TestRun.LOGGER.info(f"Writing to sectors outside of cached range.")
+        test_lba = [max_cached_lba + 1] + random.sample(
+            [
+                *range(0, min_cached_lba),
+                *range(max_cached_lba + 1, int(core.size.get_value(Unit.Blocks512)))
```
mmkayPL, line 87: shouldn't this take into account disk block size?

test/functional/tests/io_class/test_io_classification.py
jfckm: Fixed

test/functional/tests/io_class/test_io_classification.py
jfckm: I think it shouldn't, actually. I think that Linux block device addressing always is done in 512B blocks, so this should be fine.

mmkayPL (approved):

Deixx (approved):

PR #901, opened 2021-07-28
test_concurrent_flushes - use HDD disk as a core

karolinavelkaja:
Sometimes test was using nand drive for core dev (when optane was used as cache) and then test was failing because flushing was so fast. It's recommended to use HDD as a core.


Deixx (approved):
I'd go with increasing the size of the dirty data, but using slower core works too. And if at some point in the future we stop using HDDs as cores, then we can revisit this test.

jfckm (approved):

PR #910, opened 2021-07-30
[v21.3.3] Introduce lazy helper thread abstraction to avoid hung task

robertbaldyga:
Some helper threads are created at the very beginning of cache start/stop
operations, but they are used only after OCF start/stop finishes, which
may take significant amount of time. Kernel by default creates threads
that wait for the first wake up in uninterruptible state, which may trigger
hung task warning if the first wake up is called more than 120 seconds
after thread creation. To mitigate this problem we create lazy thread
abstraction that waits for a wake up in interruptible state.

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -14,6 +14,84 @@ extern u32 unaligned_io;
 extern u32 seq_cut_off_mb;
 extern u32 use_io_scheduler;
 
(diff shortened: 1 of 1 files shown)


robertbaldyga:
Retest this please

robertbaldyga:
Retest this please

mmichal10 (approved):

mmkayPL (approved):

PR #664, opened 2021-01-29
Add IOCTL tests

Ostrokrzew:
Needs https://github.com/Open-CAS/test-framework/pull/237

PR #920, opened 2021-08-12
adding API and tests for IOCTL requests (and bypassing casadm)

kajakoltan:
Initial/old reviews:
- https://github.com/Open-CAS/test-framework/pull/237 (replacing changes from test framework to OCL)
- https://github.com/Open-CAS/open-cas-linux/pull/664

### DIFF
--- a/test/functional/api/cas/ioctl/cas_structs.py
+++ b/test/functional/api/cas/ioctl/cas_structs.py
@@ -1,0 +1,108 @@
+#
+# Copyright(c) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import ctypes
+from enum import Enum
+
+
+class RequestCode(Enum):
+    START_CACHE_CODE = ctypes.c_uint(21).value
+    STOP_CACHE_CODE = ctypes.c_uint(2).value
+    SET_CACHE_STATE_CODE = ctypes.c_uint(3).value
+    INSERT_CORE_CODE = ctypes.c_uint(22).value
+    REMOVE_CORE_CODE = ctypes.c_uint(23).value
+    RESET_STATS_CODE = ctypes.c_uint(6).value
+    FLUSH_CACHE_CODE = ctypes.c_uint(9).value
+    INTERRUPT_FLUSHING_CODE = ctypes.c_uint(10).value
+    FLUSH_CORE_CODE = ctypes.c_uint(11).value
+    CACHE_INFO_CODE = ctypes.c_uint(24).value
+    CORE_INFO_CODE = ctypes.c_uint(25).value
+    PARTITION_INFO_CODE = ctypes.c_uint(14).value
+    PARTITION_SET_CODE = ctypes.c_uint(15).value
+    GET_CACHE_COUNT_CODE = ctypes.c_uint(16).value
+    LIST_CACHE_CODE = ctypes.c_uint(17).value
+    UPGRADE_CODE = ctypes.c_uint(19).value
+    GET_CORE_POOL_COUNT_CODE = ctypes.c_uint(26).value
+    GET_CORE_POOL_PATHS_CODE = ctypes.c_uint(27).value
+    CORE_POOL_REMOVE_CODE = ctypes.c_uint(28).value
+    CACHE_CHECK_DEVICE_CODE = ctypes.c_uint(29).value
+    SET_CORE_PARAM_CODE = ctypes.c_uint(30).value
+    GET_CORE_PARAM_CODE = ctypes.c_uint(31).value
+    SET_CACHE_PARAM_CODE = ctypes.c_uint(32).value
+    GET_CACHE_PARAM_CODE = ctypes.c_uint(33).value
+    GET_STATS_CODE = ctypes.c_uint(34).value
+    PURGE_CACHE_CODE = ctypes.c_uint(35).value
+    PURGE_CORE_CODE = ctypes.c_uint(36).value
+
+
+KiB = ctypes.c_ulonglong(1024).value
+MAX_STR_LEN = 4096
+MAX_ELEVATOR_NAME = 16
+
+
+class InitCache(Enum):
+    CACHE_INIT_NEW = ctypes.c_uint8(0).value
+    CACHE_INIT_LOAD = ctypes.c_uint8(1).value
+
+
+class CacheMode(Enum):
+    ocf_cache_mode_wt = ctypes.c_int(0).value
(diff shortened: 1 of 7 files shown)


test/functional/api/cas/ioctl/cas_structs.py
```diff
@@ -0,0 +1,108 @@
...
+    default = ocf_cache_mode_wt
+
+
+class CacheLineSize(Enum):
+    ocf_cache_line_size_4 = ctypes.c_ulonglong(4).value * KiB
+    ocf_cache_line_size_8 = ctypes.c_ulonglong(8).value * KiB
+    ocf_cache_line_size16 = ctypes.c_ulonglong(16).value * KiB
+    ocf_cache_line_size_32 = ctypes.c_ulonglong(32).value * KiB
+    ocf_cache_line_size_64 = ctypes.c_ulonglong(64).value * KiB
+    default = ocf_cache_line_size_4
```
robertbaldyga, line 66: Why don't we use class `Size` for proper units and then convert bytes value to `ctypes` type?

test/functional/api/cas/ioctl/ioctl.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if nr > IOC_NRMASK:
+        raise OverflowError(f"IO command value {nr} exceeds {IOC_NRMASK}")
+    nr <<= IOC_NRSHIFT
+
+    if size > IOC_SIZEMASK:
+        raise OverflowError(f"IO size value {size} exceeds {IOC_SIZEMASK}")
+    size <<= IOC_SIZESHIFT
+
+    return dir | type | nr | size
```
robertbaldyga, line 60: There are libs for that https://pypi.org/project/ioctl-opt/

test/functional/api/cas/ioctl/ioctl.py
```diff
@@ -0,0 +1,147 @@
...
+
+def send_script_with_dumped_args():
+    if not check_if_directory_exists(temp_dir):
+        create_directory(temp_dir, True)
+
+    TestRun.executor.rsync_to(script_source, script_dest)
+    chmod_numerical(script_dest, 550)
+
+    TestRun.executor.rsync_to(struct_path, struct_path)
+    chmod_numerical(struct_path, 440)
```
robertbaldyga, line 125: We finally have DUT-side scripting! :heart:
Although it may have been more generalized for use in other places (even as a part of TF).

test/functional/api/cas/ioctl/ioctl.py
```diff
@@ -0,0 +1,147 @@
...
+    with open(struct_path, 'wb') as dump_file:
+        marshal.dump(cas_ioctl_request.command_struct, dump_file)
+
+    send_script_with_dumped_args()
+    if interrupt:
+        pid = TestRun.executor.run_in_background(
+            f"{script_dest} -c {cas_ioctl_request.command} -s {struct_path}"
+        )
+        sleep(2)
+        TestRun.executor.kill_process(pid)
```
robertbaldyga, line 141: This function should at least take timeout as an argument. Otherwise it's unusable with commands like cache flush, which typically takes much longer than 2 seconds.

test/functional/api/cas/ioctl/ioctl.py
```diff
@@ -0,0 +1,147 @@
...
+    else:
+        TestRun.executor.run(f"{script_dest} -c {cas_ioctl_request.command} -s {struct_path}")
+    if check_if_directory_exists(temp_dir):
+        remove(temp_dir, True, True, True)
+    if os.path.exists(struct_path):
+        os.remove(struct_path)
```
robertbaldyga, line 147: Virtually all ioctls handled by CAS return information from kernel via ioctl structure, which means we want to have a way to receive this information back from DUT once ioctl is finished.

robertbaldyga (changes requested):

test/functional/api/cas/ioctl/cas_structs.py
```diff
@@ -0,0 +1,108 @@
...
+# Copyright(c) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import ctypes
+from enum import Enum
+
+
+class RequestCode(Enum):
+    START_CACHE_CODE = ctypes.c_uint(21).value
```
jfckm, line 11: What does this construct do? I just checked that `type(c_uint(21).value) == int` so its equivalent with `START_CACHE_CODE = 21` (maybe aside from integer wraparound effects, but that doesn't seem to be needed here)

test/functional/tests/ioctl/common_utils.py
```diff
@@ -0,0 +1,29 @@
...
+    r"Waiting for cache stop interrupted\. Stop will finish asynchronously\."
+]
+
+interrupt_start = [
+    r"Cache added successfully, but waiting interrupted\. Rollback"
+]
+
+load_and_force = [
+    r"cache\d+: Using \'force\' flag is forbidden for load operation\."
+]
```
jfckm, line 20: I don't think those messages should be in this file. Maybe more appropriate place would be in `api/cas/cli_messages.py`?

test/functional/tests/ioctl/common_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def clear_dmesg():
+    TestRun.executor.run_expect_success('dmesg -C')
+
+
+def check_dmesg(searched_phrase: str):
+    dmesg_out = TestRun.executor.run_expect_success("dmesg").stdout
+    __check_string_msg(dmesg_out, searched_phrase)
```
jfckm, line 29: Those as well, I think clearing/checking dmesg is more generic operation, not limited only to ioctl tests.

test/functional/tests/ioctl/test_ioctl_start.py
```diff
@@ -0,0 +1,113 @@
...
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+def test_ioctl_start(cache_mode, cache_line_size, load, force):
+    """
+        title: Start the cache without casadm.
+        description: |
+          Test of the ability to start the cache with IOCTL request bypassing native
+          OpenCAS manager - casadm.
+        pass_criteria:
+          - Cache started successfully without any errors if 'force' and 'load' flags are not used.
+          - When 'force' and 'load' flags are used cache is not started.
```
Deixx, line 36:
```suggestion
          - Cache started successfully without any errors if 'force' and 'load' flags are not used simultaneously.
          - When 'force' and 'load' flags are both used cache is not started.
```

PR #921, opened 2021-08-12
Remove remains of atomic writes support

mmichal10:


casadm/cas_lib.c
```diff
@@ -283,7 +283,6 @@ static struct name_to_val_mapping promotion_policy_names[] = {
 
 static struct name_to_val_mapping metadata_mode_names[] = {
 	{ .short_name = "normal", .value = CAS_METADATA_MODE_NORMAL },
-	{ .short_name = "atomic", .value = CAS_METADATA_MODE_ATOMIC },
 	{ NULL }
 };
```
robertbaldyga, line 287: Do we still need this mapping? We have only one metadata mode right now.

modules/cas_cache/volume/vol_blk_utils.c
```diff
@@ -446,23 +274,10 @@ int _cas_blk_identify_type(const char *path, uint8_t *type,
 	if (result)
 		return result;
 
-	if (*type == BLOCK_DEVICE_VOLUME) {
```
robertbaldyga, line 449: Also `NVME_CONTROLLER` type does not seem to be used anywhere. And it should no longer be considered a valid type.

robertbaldyga (changes requested):

robertbaldyga (approved):

rafalste (approved):

PR #932, opened 2021-08-20
cosmetic changes for exception logging in stop_all_caches()

karolinavelkaja:

### DIFF
--- a/test/functional/api/cas/casadm.py
+++ b/test/functional/api/cas/casadm.py
@@ -172,10 +172,11 @@ def stop_all_caches():
     if "No caches running" in list_caches().stdout:
         return
     TestRun.LOGGER.info("Stop all caches")
-    casctl_stop()
-    output = list_caches()
-    if "No caches running" not in output.stdout:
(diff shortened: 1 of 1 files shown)


test/functional/api/cas/casadm.py
```diff
@@ -172,10 +172,11 @@ def stop_all_caches():
     if "No caches running" in list_caches().stdout:
         return
     TestRun.LOGGER.info("Stop all caches")
-    casctl_stop()
-    output = list_caches()
-    if "No caches running" not in output.stdout:
-        raise CmdException("Error while stopping caches.", output)
+    stop_output = casctl_stop()
+    caches_output = list_caches()
```
robertbaldyga, line 176: We could also check exit code from those two and raise an exception accordingly,

robertbaldyga (changes requested):

test/functional/api/cas/casadm.py
Deixx: list_caches() checks its exit code

rafalste (approved):

PR #977, opened 2021-10-26
add patchset for support arm64 deb build

loyou:
with those patches, we can build open-cas-linux deb packages on arm64.


Deixx:
retest this please

rafalste:
Hi @loyou and thank you for submitting this patch!
However, we've decided to take a different approach while implementing this functionality, thus allowing user to build DEB and RPM packages as well and on all architectures instead.
Please feel free to submit additional patches or issues if this solution doesn't suit you in any way.

loyou:
Agree, PR #1001 is better.

PR #1032, opened 2021-12-27
opencas.conf supports setting policies and params

lausaa:
Now the opencas.conf supports setting cleaning and promotion policies for each cache and seq-cutoff policy for each core.


robertbaldyga:
Hi @lausaa, thank you for your patches! Currently we are under feature freeze before v21.12 release (we are accepting only fixes), and at this moment we are quite busy with testing and fixing the bugs. We will come back to your PR after v21.12 is released. Cheers!

PR #1033, opened 2021-12-28
Remove dependency on packaging Python package

jfckm:

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/open-cas-linux/1033)

### DIFF
--- a/utils/casctl
+++ b/utils/casctl
@@ -10,11 +10,13 @@ import sys
 min_ver = "3.6"
 ver = platform.python_version()
 if ver < min_ver:
-    print("Minimum required python version is {}. Detected python version is {}".format(
-                min_ver,
-                ver,
-            ),
-            file = sys.stderr)
+    print(
+        "Minimum required python version is {}. Detected python version is {}".format(
+            min_ver,
+            ver,
+        ),
+        file=sys.stderr,
+    )
     exit(1)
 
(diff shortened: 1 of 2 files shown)


utils/casctl
```diff
@@ -4,15 +4,16 @@
...
 import sys
 
-min_ver = version.parse("3.6")
-ver = version.parse(platform.python_version())
+min_ver = "3.6"
+ver = platform.python_version()
 if ver < min_ver:
-    print((f"Minimum required python version is {min_ver}\n"
-            f"Detected python version is {ver}"),
+    print("Minimum required python version is {}. Detected python version is {}".format(
```
robertbaldyga: Is there any reason to replace f-string with `format()`?

robertbaldyga (changes requested):

utils/casctl
jfckm: yup, f-strings were introduced in Python 3.6

mmichal10 (approved):

utils/casctl
robertbaldyga: and our min version, defined in the same file is also 3.6

robertbaldyga: ooh, I see

robertbaldyga (approved):

Issue #1020, opened 2021-12-14
[21.12] Calling `casadm --list` after detach doesn't produce a proper output


Deixx:
## Description

After a standby cache is detached, `casadm --list` produces an error output.

## Expected Behavior

```
# casadm -L
type    id   disk             status    write policy   device
cache   1    /dev/nvme2n1p1   Standby   -              -      # <-- maybe status Detached?
```

## Actual Behavior

```
# casadm -L
Failed to retrieve cache information
```

## Your Environment

* OpenCAS version (commit hash or tag): 21.12.0.0595.master

kmajzero:
Closing as the relater PR has been merged

PR #1043, opened 2022-01-13
Fix for casadm output when listing detached cache

kmajzero:
Fix for issue #1020 - make casadm produce a meaningful output when
listing a detached cache

### DIFF
--- a/casadm/cas_lib.c
+++ b/casadm/cas_lib.c
@@ -79,6 +79,8 @@ static const char *core_states_name[] = {
 
 #define NOT_RUNNING_STATE "Not running"
 
+#define STANDBY_DETACHED_STATE "Standby detached"
+
 #define CACHE_STATE_LENGHT 20
 
 #define CAS_LOG_FILE "/var/log/opencas.log"
@@ -487,9 +489,13 @@ void print_err(int error_code)
 		cas_printf(LOG_ERR, "%s\n", msg);
 }
 
-const char *get_cache_state_name(int cache_state)
+const char *get_cache_state_name(int cache_state, bool detached)
 {
 	int i;
+
+	if (detached == true)
+		return STANDBY_DETACHED_STATE;
+
 	/* iterate over states in reverse order, so that combined states "running&stopping"
 	 * would be described as "stopping" */
 	for (i = ocf_cache_state_max - 1; i >= 0; --i) {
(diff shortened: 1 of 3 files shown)


casadm/cas_lib.c
```diff
@@ -2792,7 +2790,7 @@ int list_caches(unsigned int list_format, bool by_id_path)
 			"%s,%u,%s,%s,%s,%s\n",
 			"cache", /* type */
 			curr_cache->id, /* id */
-			curr_cache->device, /* device path */
+			curr_cache->standby_detached ? "detached" : curr_cache->device, /* device path */
```
robertbaldyga, line 2793: We may rather want to print "-" as a device name (like we do for detached cores exp objects) and change status (`tmp_status`) so it would print "Standby" for standby attached, and "Standby detached" for detached case.

robertbaldyga (changes requested):

casadm/cas_lib.c
```diff
@@ -2779,7 +2777,14 @@ int list_caches(unsigned int list_format, bool by_id_path)
 			snprintf(mode_string, sizeof(mode_string), "wb->%s",
 					cache_mode_to_name(curr_cache->mode));
 		} else {
-			tmp_status = get_cache_state_name(curr_cache->state);
+			if (curr_cache->standby_detached) {
```
arutk, line 2780: please disable attempt to resolve by-id path earlier in "casadm -L"  in case of standby detached.

casadm/cas_lib.c
```diff
@@ -2779,7 +2777,14 @@ int list_caches(unsigned int list_format, bool by_id_path)
...
+				snprintf(status_buf, sizeof(status_buf), 
+					"Standby detached");
+				tmp_status = status_buf;
+			} else {
+				tmp_status = get_cache_state_name(curr_cache->state);
+			}
+
```
arutk, line 2787: how about adding a "detached" parameter to get_cache_state_name() and moving this entire logic to get_cache_state_name() ?

arutk (approved):

casadm/cas_lib.c
```diff
@@ -786,6 +789,7 @@ struct cache_device *get_cache_device(const struct kcas_cache_info *info, bool b
 	cache->expected_core_count = info->info.core_count;
 	cache->id = cache_id;
 	cache->state = info->info.state;
+	cache->standby_detached = info->info.standby_detached;
 
 	if (strncpy_s(cache->device, sizeof(cache->device),
 			info->cache_path_name,
```
mmichal10, line 795: What is the content of `info->cache_path_name` if cache is detached? I wonder if it wouldn't be safer to put `-` into `cache->device` instead of the unspecified(?) value.

kmajzero: info->cache_path_name is string consisting of '\0's, len = 4096

arutk (approved):

mmichal10 (approved):

PR #1053, opened 2022-01-30
Add a support for kernel 5.12

hammerg:


robertbaldyga:
ok to test

hammerg:
> ok to test

Please wait with the tests. I've found an issue with one of the configuration files and I'm reviewing my changes.

configure.d/1_bd_part_count.conf
```diff
@@ -0,0 +1,64 @@
...
+    case "$1" in
+    "1")
+		add_function "
+	static inline int cas_blk_get_part_count(struct block_device *bdev)
+	{
+		struct block_device *part;
+		unsigned long idx;
+		int count = 0;
+
+		rcu_read_lock();
```
robertbaldyga, line 33: This rcu lock doesn't seem to be necessary anymore.

configure.d/1_bd_first_part.conf
```diff
@@ -9,22 +9,44 @@
...
     "1")
 		add_function "
 	static inline int cas_bd_get_next_part(struct block_device *bd)
+	{
+		int part_no = 0;
+		struct gendisk *disk = bd->bd_disk;
+		struct block_device *part;
+		unsigned long idx;
+
+		rcu_read_lock();
```
robertbaldyga, line 37: RCU lock is not needed.  `xa_for_each()` takes RCU lock automatically.

configure.d/1_module_mutex.conf
```diff
@@ -0,0 +1,36 @@
...
+
+apply() {
+    case "$1" in
+    "1")
+        add_define "CAS_MODULE_LOCK() \\
+            mutex_lock(&module_mutex);"
+        add_define "CAS_MODULE_UNLOCK() \\
+            mutex_unlock(&module_mutex);" ;;
+    "2")
+        add_define "CAS_MODULE_LOCK()"
```
robertbaldyga, line 29: This empty define is ok because none of the kernels which have `module_mutex` hidden has `SYMBOL_LOOKUP_SUPPORTED` set. However the code would look cleaner if we also had define like `MODULE_MUTEX_SUPPORTED` and then guard dynamic symbol lookup with both `SYMBOL_LOOKUP_SUPPORTED` and `MODULE_MUTEX_SUPPORTED`. That way we can be sure that in case of unlikely even of enabling dynamic symbol lookup in one of future kernels we do not attempt to perform dynamic lookup without taking the `module_mutex`.

robertbaldyga (changes requested):

robertbaldyga:
@hammerg This is super valuable contribution, thank you!

configure.d/1_module_mutex.conf
hammerg: I think I understand, I'll submit a new patch soon. Thanks, Gal.

modules/cas_cache/main.c
```diff
@@ -83,7 +83,9 @@ int static cas_find_symbol(void *data, const char *namebuf,
 
 int static cas_casdisk_lookup_funtions(void)
 {
+#ifdef MODULE_MUTEX_SUPPORTED
```
robertbaldyga, line 86: We also need to change line 49 to
```c
#if defined(SYMBOL_LOOKUP_SUPPORTED) && defined(MODULE_MUTEX_SUPPORTED)
```
so that only if both are supported, the dynamic lookup is performed.

robertbaldyga (changes requested):

configure.d/1_module_mutex.conf
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright(c) 2012-2021 Intel Corporation
```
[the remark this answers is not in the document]
hammerg: Time goes by...

configure.d/1_module_mutex.conf
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Copyright(c) 2022 Intel Corporation
```
robertbaldyga, line 3: You may want to add Huawei copyright entry to the header in files that you added/modified.

robertbaldyga (changes requested):

robertbaldyga (approved):

mmichal10 (approved):

Issue #1046, opened 2022-01-18
[21.6.4] test_ioclass_id_as_condition: Writes were not properly cached


karolinavelkaja:
## Description

test_ioclass_id_as_condition: Writes were not properly cached

**Failing case for ext4 filesystem**

## Steps to Reproduce

1. Prepare cache and core. Disable udev.
2. Create and load IO class config file.
3. Prepare ext4 filesystem and mount /dev/cas1-1 at /tmp/cas1-1.
4. Run IO fulfilling IO class 1 condition (and not IO class 2) and check if it is classified properly.
    fio --eta=always --output-format=json --output=fio_run_2022-01-15_03-51-1642215115 --ioengine=libaio --size=1048576 --readwrite=write --filename=/tmp/cas1-1/base_dir/test_file_1 --name=fio --group_reporting
    **Failed: Writes were not properly cached!
    Expected: 1048576.0 Byte, actual: 0.0 Blocks4096**

## Logs

[2022_01_15_03_50_18_test_ioclass_id_as_condition.zip](https://github.com/Open-CAS/open-cas-linux/files/7889516/2022_01_15_03_50_18_test_ioclass_id_as_condition.zip)

## Your Environment

* OpenCAS version (commit hash or tag): 21.6.4
* Operating System: Rhel 8.4

jfckm:
The defect is caused by CAS not being able to resolve directory to inode so quickly.

Here is where it's defined in code:
https://github.com/Open-CAS/open-cas-linux/blob/7041517bf47df08b75aa4727189dc612ee8d4bc6/modules/cas_cache/classifier.c#L318-L328

As you can see CAS resolves directory by means of delayed work, which is run periodically (1s for not-yet-resolved directories, 5s for resolved). In this particular test the window between creating directory and running workload is so small that most of the times CAS doesn't have the time to resolve inode number.

Although adding slight (1s) delay before running fio is not pretty it should do the job.

karolinavelkaja:
Confirming that adding the delay resolves the issue and test is passing

karolinavelkaja:
Test fix needed, expected behavior from CAS perspective.

karolinavelkaja:
Test improvement implemented, closing.

PR #1056, opened 2022-02-03
fix for test_ioclass_id_as_condition - adding 1s delay before io

karolinavelkaja:
need to add 1s delay before io - CAS not being able to resolve directory to inode so quickly
https://github.com/Open-CAS/open-cas-linux/issues/1046

### DIFF
--- a/test/functional/tests/io_class/test_io_classification.py
+++ b/test/functional/tests/io_class/test_io_classification.py
@@ -4,6 +4,7 @@
 #
 
 import random
+import time
 from itertools import permutations
 
 import pytest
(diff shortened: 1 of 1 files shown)


test/functional/tests/io_class/test_io_classification.py
```diff
@@ -475,6 +477,7 @@ def test_ioclass_id_as_condition(filesystem):
                       "it is classified properly."):
         # Should be classified as IO class 5
         base_occupancy = cache.get_io_class_statistics(io_class_id=5).usage_stats.occupancy
+        time.sleep(1)
```
arutk, line 480: sleep(1) after IO class configure and/or directory structure change is enough

karolinavelkaja: Right! Thank you, I've tested it out and it's good:)

arutk (approved):

test/functional/tests/io_class/test_io_classification.py
```diff
@@ -444,13 +445,15 @@ def test_ioclass_id_as_condition(filesystem):
             ioclass_config_path=ioclass_config_path,
         )
         casadm.load_io_classes(cache_id=cache.cache_id, file=ioclass_config_path)
+        time.sleep(1)
```
jfckm, line 448: I think that it might be beneficial to put a comment in here why test sleeps

Deixx: also instead of an arbitrary value we have `MAX_CLASSIFICATION_DELAY = timedelta(seconds=6)` in ioclass_config.py

mmichal10 (approved):

Issue #1019, opened 2021-12-13
[21.12] Standby cache activated without access to backend device


Deixx:
## Description

After main node power failure and shutting down core DRBD device standby cache was activated (incomplete state).

## Expected Behavior

Activation blocked

## Actual Behavior

Standby cache activated in an incomplete state

## Your Environment

* OpenCAS version (commit hash or tag): 21.12.0.0595.master
* Operating System: RHEL 8.3

Deixx:
Discuss if this should be expected behavior

Deixx:
Config improvements - load standby cache flag

PR #1057, opened 2022-02-04
Standby load config

jfckm:
Support standby load in opencas.conf + some minor fixes around that

---
This change is [Reviewable](https://reviewable.io/reviews/open-cas/open-cas-linux/1057)


robertbaldyga:
retest this please

modules/cas_cache/layer_cache_management.c
```diff
@@ -2385,11 +2397,31 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
...
 		if (result) {
 			module_put(THIS_MODULE);
 			return result;
 		}
+
+		result = ocf_mngt_cache_get_by_name(cas_ctx, cache_name_meta,
+				OCF_CACHE_NAME_SIZE, &tmp_cache);
+		/* Need to return name from metadata now for caller to properly
+		 * communicate the error to user */
+		BUG_ON(cache_id_from_name(&cmd->cache_id, cache_name_meta));
```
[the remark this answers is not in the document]
arutk: I would move cache_id_from_name() call to line 2407 - now it is located strangely right between setting the result and checking it.

modules/cas_cache/layer_cache_management.c
```diff
@@ -2385,11 +2397,31 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
...
+
+		if (result != -OCF_ERR_CACHE_NOT_EXIST) {
```
arutk, line 2414: I think any other result != 0 should also be treated as error

modules/cas_cache/layer_cache_management.c
```diff
@@ -2385,11 +2397,31 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
...
+			printk(KERN_ERR "Can't load %s. Cache using that name "
+					"already exists.\n", cache_name_meta);
+			module_put(THIS_MODULE);
+			return -OCF_ERR_CACHE_EXIST;
+		}
+
```
arutk, line 2420: missing ocf_mngt_cache_put

utils/opencas.conf.5
```diff
@@ -24,7 +24,7 @@ Cache device <DEVICE>
 .br
 Cache mode {wt|wb|wa|pt|wo}
 .br
-Extra fields (optional) ioclass_file=<file>,cleaning_policy=<alru,nop>,promotion_policy=<always,nhit>
+Extra fields (optional) ioclass_file=<file>,cleaning_policy=<alru,nop>,promotion_policy=<always,nhit>,target_failover_state=<active,standby>
```
arutk, line 27: is there any difference between specifying "target_failover_state=active" and not specifying target_failover_state at all?

arutk (changes requested):

utils/opencas.conf.5
jfckm: No, there's none. However, all the options are specified in key=value format and I find `target_failover_state=<standby>` strange

modules/cas_cache/layer_cache_management.c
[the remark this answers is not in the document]
jfckm: RE BUG_ON: I think that it's apropriate - OCF returned no error, so I'm assuming all data retrieved from metadata is correct, if it's not then we're in trouble

modules/cas_cache/layer_cache_management.c
jfckm: This check verifies if there's already any cache started and running with the same name that we're trying to load to avoid name collision. The error condition is here everything that isn't `OCF_ERR_CACHE_NOT_EXIST`.

robertbaldyga (changes requested):

modules/cas_cache/layer_cache_management.c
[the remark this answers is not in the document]
robertbaldyga: With this BUG_ON it's possible to prepare metadata on drive (actually just change cache name and recalculate CRC, so nothing sophisticated) so that loading in causes kernel panic. In case of cache name it's even worse, because malicious intervention is not even needed - we have OCF in SPDK, and it can produce metadata with cache names not compatible with OCL convention. Please add proper error handling.

robertbaldyga (changes requested):

modules/cas_cache/layer_cache_management.c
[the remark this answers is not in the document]
jfckm: That's right, I've fixed it and moved the call earlier.

casadm/cas_lib.c
```diff
@@ -998,9 +998,12 @@ int start_cache(uint16_t cache_id, unsigned int cache_init,
 	cmd.line_size = line_size;
 	cmd.force = (uint8_t)force;
 
-	if (run_ioctl_interruptible_retry(fd, KCAS_IOCTL_START_CACHE, &cmd,
-			"Starting cache", cache_id, OCF_CORE_ID_INVALID) < 0) {
+	status = run_ioctl_interruptible_retry(fd, KCAS_IOCTL_START_CACHE, &cmd,
+			"Starting cache", cache_id, OCF_CORE_ID_INVALID);
+	cache_id = cmd.cache_id;
```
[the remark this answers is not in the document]
jfckm: Done

modules/cas_cache/layer_cache_management.c
jfckm: Fixed

test/functional/api/cas/cli_messages.py
```diff
@@ -81,7 +81,7 @@
 ]
 
 load_and_force = [
-    r"Use of \'load\' and \'force\' simultaneously is forbidden\."
```
robertbaldyga, line 84: This output is no longer accurate after recent changes.

robertbaldyga (changes requested):
Please also update the documentation https://open-cas.github.io/guide_configuring.html

jfckm:
Fixes #1019

casadm/cas_main.c
```diff
@@ -406,17 +406,24 @@ int handle_start()
...
-	}
+	if (command_args_values.state == CACHE_INIT_LOAD) {
+		if (command_args_values.force ||
+			 command_args_values.line_size != ocf_cache_line_size_none ||
+			 command_args_values.cache_mode != ocf_cache_mode_none ||
+			 command_args_values.cache_id != OCF_CACHE_ID_INVALID) {
+			cas_printf(LOG_ERR, "Use of 'load' with 'force', 'cache-id',"
+					" 'cache-mode' or 'cache-line-size'"
+					" simultaneously is forbidden.\n");
+			return FAILURE;
```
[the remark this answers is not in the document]
jfckm: done

arutk (approved):

PR #1074, opened 2022-02-25
Casadm error message update

kmajzero:
Error message updated for the case of not meeting the space requirements
for the cache device.

### DIFF
--- a/casadm/extended_err_msg.c
+++ b/casadm/extended_err_msg.c
@@ -164,7 +164,7 @@ struct {
 	},
 	{
 		OCF_ERR_INVAL_CACHE_DEV,
-		"Device does not meet requirements."
+		"Device does not meet requirements. See dmesg for more information"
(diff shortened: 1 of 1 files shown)


casadm/extended_err_msg.c
```diff
@@ -164,7 +164,7 @@ struct {
 	},
 	{
 		OCF_ERR_INVAL_CACHE_DEV,
-		"Device does not meet requirements."
+		"Device does not meet requirements. Check kernel dmesg for details"
```
jfckm, line 167: It would be good if we used same wording in both `KCAS_ERR_DETACHED` and here - they both direct users to check dmesg.

jfckm (changes requested):

jfckm (approved):

robertbaldyga (approved):

PR #1082, opened 2022-03-07
initializing completion twice replaced with reinit

pdebski21:
According to documentation using init_completion() twice is potential bug and use of reinit_completion() is suggested.

https://www.kernel.org/doc/Documentation/scheduler/completion.txt

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -1,5 +1,5 @@
 /*
-* Copyright(c) 2012-2021 Intel Corporation
+* Copyright(c) 2012-2022 Intel Corporation
 * SPDX-License-Identifier: BSD-3-Clause
 */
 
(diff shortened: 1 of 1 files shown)


mmichal10 (approved):

modules/cas_cache/layer_cache_management.c
```diff
@@ -180,7 +180,7 @@ static int _cache_mngt_lock_sync(ocf_cache_t cache)
 	if (!context)
 		return -ENOMEM;
 
-	_cache_mngt_async_context_init(context);
```
robertbaldyga, line 183: Maybe we could introduce more descriptive pair of `_cache_mngt_async_context_init()` / `_cache_mngt_async_context_reinit()`. To me it looks more consise and informative than additional boolean parameter.

robertbaldyga (changes requested):

modules/cas_cache/layer_cache_management.c
pdebski21: I was considering it but I was not sure which one will be better. Your suggestion assures me that adding `_cache_mngt_async_context_reinit()` function is better solution.

modules/cas_cache/layer_cache_management.c
```diff
@@ -148,15 +148,28 @@ static int _cache_mngt_async_caller_set_result(
 	return result;
 }
 
-static inline void _cache_mngt_async_context_init(
+static inline void _cache_mngt_async_context_common(
```
robertbaldyga, line 151: I'd call it `_cache_mngt_async_context_init_common()`, because right now it's not very descriptive.

robertbaldyga (changes requested):

robertbaldyga:
You also need to update the date in the file header.

robertbaldyga (approved):

arutk (approved):

PR #1087, opened 2022-03-08
Error message fix for OCL issue 1065

kmajzero:
Error message changed not to suggest --force option


casadm/extended_err_msg.c
```diff
@@ -231,8 +231,7 @@ struct {
 	},
 	{
 		KCAS_ERR_CONTAINS_PART,
-		"Device contains partitions.\nIf you want to continue, "
-		"please use --force option.\nWarning: all data will be lost!"
+		"Operation not allowed. Device contains partitions."
```
robertbaldyga, line 234: This breaks message printed in scenarios other than `--standby --activate`. We want to change message only in this specific scenario, keeping it as it was for other ones.

robertbaldyga (changes requested):

casadm/extended_err_msg.c
kmajzero: Ok, custom message will be printed by casadm for this case

kmajzero:
Closing, as Robert created better fix for this issue in PR #1088

PR #1090, opened 2022-03-11
tests: fix capabilities test

### LINKED ISSUE #880: Cas device max_hw_sectors_kb is not set properly
## Description

Issue during test_device_capalilities - Cas device max_hw_sectors_kb is not set properly
Actual value should be greater:
Cas device max_hw_sectors_kb is not set properly. Is: 2147483646, should be 2147483647 (cache: 128, core: 2147483647)

Issue occurs for the step/properties: "SCSI-debug module | dev_size_mb = 2048 | logical_block_size = 2048 | max_sectors_kb = 1024"

## Steps to Reproduce

Test body 
1. SCSI-debug module | dev_size_mb = 1024 | logical_block_size = 512 | max_sectors_kb = 1024
2. SCSI-debug module | dev_size_mb = 1024 | logical_block_size = 512 | max_sectors_kb = 256
3. SCSI-debug module | dev_size_mb = 1024 | logical_block_size = 512 | max_sectors_kb = 128
4. **SCSI-debug module | dev_size_mb = 2048 | logical_block_size = 2048 | max_sectors_kb = 1024**
        1) Prepare devices.
        2) Start cache and add prepared core device as core.
        3) Compare capabilities for CAS device, cache and core (or check proper error if logical sector mismatch occurs).
            Cas device logical_block_size has proper value: 2048 (cache: 512, core: 2048)
            **Cas device max_hw_sectors_kb is not set properly. Is: 2147483646, should be 2147483647 (cache: 128, core: 2147483647)**

## Your Environment

* OpenCAS version (commit hash or tag): 21.06.0.0530.master
* Operating System: Centos 5.9
* Kernel version: 5.10
* Cache device type (NAND/Optane/other):
* Core device type (HDD/SSD/other):
* Cache configuration:
* * Cache mode: (default: wt)
* * Cache line size: (default: 4)
* * Promotion policy: (default: always)
* * Cleaning policy: (default: alru)
* * Sequential cutoff policy: (default: full)
* Other (e.g. _lsblk_, _casadm -P_, _casadm -L_)

mmichal10:
Adjust manner of calculating some of the traits to the newer kernel version.

New way of calculating them have been introduced in kernel with commit
97f433c3601a24d3513d06f575a389a2ca4e11e4.

Closes #880 

### DIFF
--- a/test/functional/tests/misc/test_device_capabilities.py
+++ b/test/functional/tests/misc/test_device_capabilities.py
@@ -183,11 +183,23 @@ def compare_capabilities(cache_device, core_device, cache, core, msg):
             cache_val = cache_dev_capabilities[capability]
             core_val = core_dev_capabilities[capability]
 
-            comparison_val = method(core_val, cache_val) if method is not None else core_val
+            expected_val = method(core_val, cache_val) if method is not None else core_val
 
-            if comparison_val != cas_val:
+            if capability in ["max_sectors_kb", "max_hw_sectors_kb"] and expected_val != cas_val:
+                # On the newer kernels this trait is rounded. Instead of checking for
+                # the current kernel version, assume that both values are acceptable.
+                SECTOR_SHIFT = 9
+                lbs = measure_capabilities(core)["logical_block_size"]
+                # The original uint is kb, but number of sectors is needed
+                new_expected_val = expected_val * 2
+                round_val = lbs >> SECTOR_SHIFT
+                new_expected_val -= new_expected_val % round_val
+                # Restore the original unit
+                expected_val = new_expected_val // 2
+
+            if expected_val != cas_val:
                 TestRun.LOGGER.error(f"Cas device {capability} is not set properly. Is: {cas_val}, "
-                                     f"should be {comparison_val} (cache: {cache_val}, "
+                                     f"should be {expected_val} (cache: {cache_val}, "
                                      f"core: {core_val})")
                 continue
             TestRun.LOGGER.info(f"Cas device {capability} has proper value: {cas_val} "


test/functional/tests/misc/test_device_capabilities.py
```diff
@@ -183,11 +183,23 @@ def compare_capabilities(cache_device, core_device, cache, core, msg):
...
             core_val = core_dev_capabilities[capability]
 
-            comparison_val = method(core_val, cache_val) if method is not None else core_val
-
-            if comparison_val != cas_val:
+            expected_val = method(core_val, cache_val) if method is not None else core_val
+
+            if capability in ["max_sectors_kb", "max_hw_sectors_kb"] and expected_val != cas_val:
+                # On the newer kernels this trait is rounded. Instead of checking for
+                # the current kernel version, assume that both values are acceptable.
```
rafalste, line 190: Wouldn't it introduce the risk for a pass in case of a justified mismatch? What I mean is - is it possible that _cas_val_ will be set to wrong value, which is equal to rounded _expected_val_?

karolinavelkaja (approved):

test/functional/tests/misc/test_device_capabilities.py
mmichal10: Yes, I'm aware of that but I consider it highly unlikely. Additionally, I believe this test main purpose is to verify if traits are inherited properly (either from cache or core device, depend on a particular trait), not to check the values themselves. Also, this logic with accepting two values may be removed as soon as we drop support for kernels below 5.10 :smile:

rafalste (approved):

PR #1115, opened 2022-03-22
Add workflow with copyright date verification

ArkadiuszNeumann:


.github/verify_header.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+YEAR=$(date +"%Y")
+REGEX="(Copyright\(c\) )([0-9]{4}-([0-9]{4}))( Intel Corporation)"
```
arutk, line 4: not necessarily Intel Corporation - there may be multiple lines with different companies

PR #1116, opened 2022-03-22
Fix error message about not enough amount of RAM

### LINKED ISSUE #1100: Unable to start cache due to not enaugh free RAM - missleading error message
## Description
Starting cache on pretty big 7T nvme drive error message is missleading and points to too small disk size or cache line size.

## Actual behaviour

```
[root@gklab-109-055 ~]# lsblk
NAME        MAJ:MIN RM    SIZE RO TYPE MOUNTPOINT
loop0         7:0    0    8.8G  1 loop /DVD
sda           8:0    0  223.6G  0 disk 
|-sda1        8:1    0      1G  0 part /boot/efi
|-sda2        8:2    0     10G  0 part [SWAP]
`-sda3        8:3    0    100G  0 part /
sdb           8:16   0  931.5G  0 disk 
sdc           8:32   0  931.5G  0 disk 
sdd           8:48   0  931.5G  0 disk 
sde           8:64   0  931.5G  0 disk 
nvme1n1     259:0    0  349.3G  0 disk 
`-nvme1n1p1 259:1    0 1023.8M  0 part 
nvme0n1     259:2    0      7T  0 disk 

[root@gklab-109-055 ~]# casadm -S -d `dev nvme0n1`
Not enough free RAM.
You need at least 0GB to start cache with cache line size equal 4kB.
Try with greater cache line size.
```
## Proposed behaviour
based on dmesg log error message should inform that there is no enough free RAM for cache metadata to start cache
and good approach will be attach available and needed RAM

## Logs
dmesg:
```
[82044.309325] cache1: Not enough free RAM for cache metadata to start cache
[82044.316927] cache1: Available RAM: 95853039616 B
[82044.323144] cache1: Needed RAM: 144518562782 B
[82044.328476] cache1: Attaching cache device failed
```

mmichal10:
Closes #1100 

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -2540,7 +2540,6 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
 
 	strlcpy(cmd->cache_elevator, context->cache_elevator,
 			MAX_ELEVATOR_NAME);
-	cmd->min_free_ram = context->min_free_ram;
 
 	result = _cache_start_finalize(cache, cmd->init_cache, false);
 	if (result)
@@ -2555,6 +2554,8 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
 
 	return result;
 err:
+	cmd->min_free_ram = context->min_free_ram;
+
 	_cache_mngt_async_context_reinit(&context->async);
 	ocf_mngt_cache_stop(cache, _cache_mngt_cache_stop_rollback_complete,
 			context);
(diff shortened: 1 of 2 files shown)


modules/cas_cache/layer_cache_management.c
```diff
@@ -2555,6 +2554,8 @@ int cache_mngt_init_instance(struct ocf_mngt_cache_config *cfg,
 
 	return result;
 err:
+	cmd->min_free_ram = context->min_free_ram;
```
jfckm, line 2557: This value won't be set before line 2511, so we might be leaking 8B of uninitialized kernel memory here.

jfckm (changes requested):

modules/cas_cache/layer_cache_management.c
mmichal10: `context` is allocated with `kzalloc()` so it is initialized

jfckm: right, I must've overlooked that

jfckm (approved):

arutk (approved):

PR #1120, opened 2022-03-22
Add basic workflow with copyright date verification

ArkadiuszNeumann:


rafalste (commented):
Hi @ArkadiuszNeumann,
I have some suggestions here to simplify the process and make it a bit more efficient. And combining all of my comments together, what do you think to rewrite _verify_header.sh_ it like this:
```
YEAR=$(date +"%Y")
REGEX="Copyright\(c\) [0-9]{4}-([0-9]{4}) |Copyright\(c\) ([0-9]{4}) "

while read -r line; do
    if [[ "$line" =~ $REGEX ]]; then
        echo ${BASH_REMATCH[0]}
        if [[ $YEAR == ${BASH_REMATCH[1]} || $YEAR == ${BASH_REMATCH[2]} ]]; then
            echo $1 have appropriate license header
            exit 0
        fi
        echo $1 have wrong license header year
        exit 1
    fi
done < "$1"

echo $1 does not contain appropriate license header
exit 1
```

ArkadiuszNeumann:
Hi @rafalste 
ready for review

.github/verify_header.sh
```diff
@@ -0,0 +1,23 @@
...
+        echo ${BASH_REMATCH[0]}
+        if [[ $YEAR == ${BASH_REMATCH[1]} || $YEAR == ${BASH_REMATCH[2]} ]]; then
+            echo $1 have appropriate license header
+            exit 0
+        fi
+        continue
+    fi
+done < "$1"
+
+ echo $1 have wrong license header year
```
rafalste, line 22: I'm just wondering why you moved this message outside the loop. In current implementation we don't actually know what went wrong here - if just the year was not set properly, or if the license header haven't been found at all. In the code that I suggested, this message was in place where `continue` is now, and that way we knew that header was found but with a wrong year.

rafalste (commented):
Hey @ArkadiuszNeumann and thank you for a quick response. :)
Please see my further comments.

.github/verify_header.sh
```diff
@@ -0,0 +1,23 @@
...
+REGEX="Copyright\(c\) [0-9]{4}-([0-9]{4}) |Copyright\(c\) ([0-9]{4}) "
+
+while read -r line; do
+    if [[ "$line" =~ $REGEX ]]; then
+        echo ${BASH_REMATCH[0]}
+        if [[ $YEAR == ${BASH_REMATCH[1]} || $YEAR == ${BASH_REMATCH[2]} ]]; then
+            echo $1 have appropriate license header
+            exit 0
+        fi
+        continue
```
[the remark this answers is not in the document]
ArkadiuszNeumann: right, continue is useless here

.github/verify_header.sh
```diff
@@ -0,0 +1,23 @@
...
+REGEX="Copyright\(c\) [0-9]{4}-([0-9]{4}) |Copyright\(c\) ([0-9]{4}) "
+
+while read -r line; do
+    if [[ "$line" =~ $REGEX ]]; then
+        echo ${BASH_REMATCH[0]}
+        if [[ $YEAR == ${BASH_REMATCH[1]} || $YEAR == ${BASH_REMATCH[2]} ]]; then
+            echo $1 have appropriate license header
+            exit 0
+        fi
+        echo $1 have wrong license header year
```
rafalste, line 18: \+ 'exit 1' after wrong header year was found

rafalste (commented):
Also please add `-s` to `git commit` command to sign off your commit. You will also need to set your username and email in git config if you didn't do it before:
```
git config --global user.name "Name Surname"
git config --global user.email address@intel.com
```

.github/verify_header.sh
ArkadiuszNeumann: But if there are for example few headers then it will exit after first wrong match. 
Without  this 'exit 1' it will test every header, so if someone from other company will add his own header it will test it.

KamilLepek:
Hi, @ArkadiuszNeumann is no longer working on this feature. I'll finish it next week.

KamilLepek:
Will be moved to another PR

PR #1127, opened 2022-03-23
Don't allow flush/purge while cache is in standby

### LINKED ISSUE #1117: Invalid management operation does not provide error message (Purge cache)
## Description
Not allowed in standby state management operation does not provide error message

## Expected Behavior
[root@localhost open-cas-linux]# casadm --script --purge-cache --cache-id 1
Cannot purge cache while cache 1 is in standby mode

## Actual Behavior
[root@localhost open-cas-linux]# casadm --script --purge-cache --cache-id 1

jfckm:
Fixes #1117 

This change is [Reviewable](https://reviewable.io/reviews/open-cas/open-cas-linux/1127)

### DIFF
--- a/modules/cas_cache/layer_cache_management.c
+++ b/modules/cas_cache/layer_cache_management.c
@@ -880,6 +880,11 @@ int cache_mngt_flush_device(const char *cache_name, size_t name_len)
 	if (result)
 		return result;
 
+	if (ocf_cache_is_standby(cache)) {
+		ocf_mngt_cache_put(cache);
+		return -OCF_ERR_CACHE_STANDBY;
+	}
+
 	result = _cache_mngt_read_lock_sync(cache);
 	if (result) {
 		ocf_mngt_cache_put(cache);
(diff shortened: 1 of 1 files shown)


modules/cas_cache/layer_cache_management.c
```diff
@@ -854,6 +854,11 @@ int cache_mngt_purge_device(const char *cache_name, size_t name_len)
 	if (result)
 		return result;
 
+	if (ocf_cache_is_standby(cache)) {
```
robertbaldyga, line 857: Don't we want to handle this in OCF? IMO this check doesn't belong to the adapter layer.

modules/cas_cache/layer_cache_management.c
```diff
@@ -875,6 +880,11 @@ int cache_mngt_flush_device(const char *cache_name, size_t name_len)
...
```
robertbaldyga, line 883: Ditto.

robertbaldyga (changes requested):

jfckm:
Obsolete. Fixed in https://github.com/Open-CAS/ocf/pull/678

PR #1134, opened 2022-03-24
Packages and permissions improvements

rafalste:
- fix files/dirs permissions
- ddiskit related modifications
- improve RPM/DEB package creation process
- clean a bit (missing licenses, dead links, etc.)


modules/Makefile
```diff
@@ -26,7 +26,7 @@ OCFDIR=$(PWD)/../ocf
 PWD=$(shell pwd)
 KERNEL_VERSION ?= "$(shell uname -r)"
 KERNEL_DIR ?= "/lib/modules/$(KERNEL_VERSION)/build"
-MODULES_DIR=/lib/modules/$(shell uname -r)/extra
+MODULES_DIR=/lib/modules/$(shell uname -r)/extra/block/opencas
```
arutk, line 29: use $KERNEL_VERSION instead of hardcoding "uname -r"

tools/pckgen.d/rpm/CAS_NAME.spec
```diff
@@ -108,16 +108,16 @@ if [ $1 -eq 0 ]; then
     . /etc/os-release
     if [[ ! "$ID_LIKE" =~ suse|sles ]]; then
         # Search for all CAS modules to remove them from weak-modules
-        find /lib/modules/*/extra/ -name "cas_*.ko" >/var/run/rpm-open-cas-linux-modules
+        # Use realpath to resolve any possible symlinks (needed for weak-modules)
+        realpath $(find /lib/modules/*/extra/ -name "cas_*.ko") >/var/run/rpm-open-cas-linux-modules
```
arutk, line 112: we know exactly the location of the files that we installed:
/lib/modules/%{kver}/extra/block/opencas
Searching just everywhere (/lib/modules/*/extra) seems unnecessary + would break if there is more than 1 cas modules package installed

arutk (changes requested):

modules/Makefile
rafalste: done

tools/pckgen.d/rpm/CAS_NAME.spec
rafalste: While I can add the 'block/opencas' part to the search path to virtually narrow the search, I can't change this bash glob (*) to RPM macro (%{kver}), because those macros are resolved during RPM build time and not during package install/uninstall (this was one of the most inconvenient things to work with during preparing this spec).
So, if you build your package on one kernel and install it on different but still compatible kernel, this scriptlet will search for modules in the 'build environment' kernel directory, and will fail to remove them as well as remove links created by weak-modules script.

rafalste: But on the other hand... modules will also be installed in directories from build environment as stated in Makefile. So, maybe replacing this glob with the exact path won't be a problem here. Let me do a little further investigation.

arutk: ok, thanks for explanation

arutk (approved):

jfckm (approved):

tools/pckgen.d/rpm/CAS_NAME.spec
rafalste: Ąfter a quick investigation, I think we may change this glob to the exact path to modules, as this step is needed only for removing symlinks previously created by weak-modules, and in fact, those modules will be installed in the fixed 'build time' kernel path anyway. But I will make those amendments in the separate patch, after I test those to make sure it will not brake the fragile weak-modules mechanism.

robertbaldyga (approved):

PR #1148, opened 2022-03-30
Add 'wi' (Write-Invalidate) as valid cache mode

andreatomassetti:
When enabling experimental Write-Invalidate cache mode, `casctl init` fails without this patch.

Signed-off-by: Andrea Tomassetti <andrea.tomassetti@devo.com>

### DIFF
--- a/utils/opencas.py
+++ b/utils/opencas.py
@@ -292,7 +292,7 @@ class cas_config(object):
                 )
 
         def check_cache_mode_valid(self, cache_mode):
-            if cache_mode not in ['wt', 'pt', 'wa', 'wb', 'wo']:
+            if cache_mode not in ['wt', 'pt', 'wa', 'wb', 'wo', 'wi']:
                 raise ValueError(f'Invalid cache mode {cache_mode}')
 
         def check_cleaning_policy_valid(self, cleaning_policy):


bkarpins:
retest this please

utils/opencas.py
```diff
@@ -292,7 +292,7 @@ def check_cache_device_empty(self):
                 )
 
         def check_cache_mode_valid(self, cache_mode):
-            if cache_mode not in ['wt', 'pt', 'wa', 'wb', 'wo']:
+            if cache_mode not in ['wt', 'pt', 'wa', 'wb', 'wo', 'wi']:
```
robertbaldyga, line 295: On the other hand when experimental features are disabled (which is default), WI should not be recognized as a valid cache mode. This option should be accepted only conditionally. The problem is that we currently do not have any good way to recognize whether OpenCAS was compiled with experimental features enabled, so this is another thing to be implemented in order to make `casctl` work with Write-Invalidate.

robertbaldyga (changes requested):

utils/opencas.py
andreatomassetti: Shouldn't, at some point, some C function fails if WI was not enabled during compile time? Making this patch "safe" even when WI is disabled? For example:
```
inline int validate_str_cache_mode(const char *s)
```

PR #1167, opened 2022-04-06
Test standby activate with negative cli parameters

pdebski21:

### DIFF
--- a/test/functional/api/cas/cache.py
+++ b/test/functional/api/cas/cache.py
@@ -173,3 +173,6 @@ class Cache:
         return CacheConfig(self.get_cache_line_size(),
                            self.get_cache_mode(),
                            self.get_cleaning_policy())
+
+    def standby_detach(self, shortcut: bool = False):
+        return casadm.standby_detach_cache(cache_id=self.cache_id, shortcut=shortcut)
--- a/test/functional/tests/cli/test_cli_standby.py
+++ b/test/functional/tests/cli/test_cli_standby.py
@@ -79,3 +79,83 @@ def test_standby_neg_cli_params():
                     f'Expected error message in format "{disallowed_param[0]}" '
                     f'Got "{output.stderr}" instead.'
                 )
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+def test_activate_neg_cli_params():
+    """
+    title: Verifying parameters for activating a standby cache instance.
+    description: |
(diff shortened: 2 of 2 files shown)


test/functional/tests/cli/test_cli_standby.py
```diff
@@ -0,0 +1,98 @@
...
+        cache_id = 1
+        cache_line_size = 32
+
+    with TestRun.step("Init standby cache"):
+        cmd_init = casadm_bin + f"--standby --init --cache-id {cache_id} " \
+                                f"--cache-device {cache_device.path} --cache-line-size" \
+                                f" {cache_line_size} --force"
+        output = TestRun.executor.run(cmd_init)
+        if output.exit_code == 0:
+            TestRun.LOGGER.error(f'"Standby activate failed with command {cmd_init}"')
```
mmichal10, line 39: Consider using `standby_init()` function from `api/cas/casadm.py`. It wraps setting up a command and handling error. Additionally it returns `Cache` object.

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -0,0 +1,98 @@
...
+
+    with TestRun.step("Detach standby cache"):
+        cmd_init = casadm_bin + f"--standby --detach --cache-id {cache_id}"
+        output = TestRun.executor.run(cmd_init)
+        if output.exit_code == 0:
+            TestRun.LOGGER.error(f'"Standby activate failed with command {cmd_init}"')
```
mmichal10, line 45: How about adding `standby_detach()` method to `Cache` class?

test/functional/api/cas/cache.py
```diff
@@ -173,3 +173,6 @@ def get_cache_config(self):
         return CacheConfig(self.get_cache_line_size(),
                            self.get_cache_mode(),
                            self.get_cleaning_policy())
+
+    def standby_detach(self, shortcut: bool = False):
+        casadm.standby_detach_cache(cache_id=self.cache_id, shortcut=shortcut)
```
mmichal10, line 178: `casadm.standby_detach_cache()` returns an object with result, stdout and stderr. You need to pass it to the caller.

pdebski21: Adding return will be enaugh? I don't see value in using it in this specific test.

mmichal10: I agree it is unnecessary in this test, but in general API should be consistent

mmichal10 (approved):

PR #1184, opened 2022-04-20
Test standby neg cache

mmichal10:


test/functional/tests/fault_injection/test_standby.py
```diff
@@ -0,0 +1,84 @@
...
+      - The activation is cancelled
+      - The cache remains in Standby detached state after an unsuccessful activation
+      - The cache exported object is present after an unsuccessful activation
+      - A proper error message is displayed
+    """
+    with TestRun.step("Prepare two partitions of the same size"):
+        disk = TestRun.disks["cache_1"]
+        disk.create_partitions([Size(200, Unit.MebiByte), Size(200, Unit.MebiByte)])
+        active_dev = disk.partitions[0]
+        passive_dev = disk.partitions[1]
```
pdebski21, line 35: I'm not sure if passive is best description. I wonder if we should use active and standby names in all tests.
What do you think?

pdebski21 (approved):

karolinavelkaja (approved):

PR #1193, opened 2022-04-27
Example test using DRBD API

KlaudiaJ:
Extend example_test.py with test case regarding DRBD API.

### DIFF
--- a/test/functional/tests/example/example_test.py
+++ b/test/functional/tests/example/example_test.py
@@ -8,8 +8,10 @@ import pytest
 from core.test_run import TestRun
 from storage_devices.disk import DiskType, DiskTypeSet
 from storage_devices.raid import Raid, RaidConfiguration, MetadataVariant, Level
+from storage_devices.drbd import Resource, Node, install_drbd
 from test_tools import fs_utils
 from test_tools.disk_utils import Filesystem
+from test_tools.drbdadm import Drbdadm
 from test_utils.filesystem.directory import Directory
 from test_utils.filesystem.file import File
 from test_utils.size import Size, Unit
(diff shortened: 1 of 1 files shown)


test/functional/tests/example/example_test.py
```diff
@@ -147,3 +149,62 @@ def test_example_multidut():
...
+                Drbdadm.create_md(caches.name)
+                Drbdadm.up(caches.name)
+
+    with TestRun.step("Set DUT 1 as primary"):
+        with TestRun.use_dut(dut1):
+            Drbdadm.primary(caches.name, True)  # use --force only for initial synchronization
+
+    with TestRun.step("Set DUT 2 as secondary"):
+        with TestRun.use_dut(dut2):
+            Drbdadm.secondary(caches.name)
```
mmichal10, line 196: Is it required to set the second instance as secondary if the primary node has been set in the previous step?

mmichal10 (approved):

test/functional/tests/example/example_test.py
```diff
@@ -147,3 +149,62 @@ def test_example_multidut():
                 TestRun.LOGGER.info(dut1_ex.run_expect_success("which casctl").stdout)
                 for name, disk in TestRun.disks.items():
                     TestRun.LOGGER.info(f"{name}: {disk.path}")
+
+
+@pytest.mark.multidut(2)
+def test_drbd_example():
```
mmichal10, line 155: Test is missing a fixture about required disks

test/functional/tests/example/example_test.py
```diff
@@ -147,3 +149,62 @@ def test_example_multidut():
...
+        install_drbd()
+        nodes = []
+        for dut in TestRun.duts:
+            with TestRun.use_dut(dut):
+                TestRun.dut.hostname = TestRun.executor.run_expect_success('uname -n').stdout
+                TestRun.LOGGER.info(TestRun.dut.hostname)
+                for name, disk in TestRun.disks.items():
+                    TestRun.LOGGER.info(f"{name}: {disk.path}")
+                # TODO: retrieve disk path (/dev/nvme2n1) from config file?
+                nodes.append(Node(TestRun.dut.hostname, "/dev/nvme2n1", "/dev/sdb", dut.ip, "7790"))
```
mmichal10, line 174: You have to use `disk.path` instead of string literals

mmichal10 (changes requested):

test/functional/tests/example/example_test.py
mmichal10: I suggest using ram disks for drbd metadata. I believe we have ram disk api integrated in test-framework already.

PR #1196, opened 2022-04-28
Casctl: Python version check fix

kmajzero:
Fixes the python version check code in the casctl.

### DIFF
--- a/utils/casctl
+++ b/utils/casctl
@@ -3,20 +3,12 @@
 # Copyright(c) 2012-2021 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
-
-import platform
 import sys
 
-min_ver = "3.6"
-ver = platform.python_version()
-if ver < min_ver:
-    print(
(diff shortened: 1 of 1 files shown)


mmichal10 (approved):

utils/casctl
```diff
@@ -3,17 +3,12 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
 #
-
-import platform
 import sys
 
-min_ver = "3.6"
-ver = platform.python_version()
-if ver < min_ver:
+if sys.version_info < (3, 6):
```
rafalste, line 8: I would export the minimum version number to a separate variable (just as it was before), to have it in one place instead of multiple.
I don't know how much value it has in this case (since there are only two of them and they are pretty close to each other), but this condition could look like this:
`if sys.version_info < tuple(map(int, min_ver.split('.'))):`

rafalste (approved):

robertbaldyga (approved):

rafalste (approved):

PR #1200, opened 2022-05-05
Test activate without detach

pdebski21:

### DIFF
--- a/test/functional/tests/cli/test_cli_standby.py
+++ b/test/functional/tests/cli/test_cli_standby.py
@@ -5,8 +5,8 @@
 
 import pytest
 
-from api.cas import casadm
-from api.cas.cli import casadm_bin
+from api.cas import casadm, casadm_parser
+from api.cas.cli import casadm_bin, standby_activate_cmd
 from core.test_run import TestRun
 from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
 from test_utils.size import Size, Unit
@@ -16,9 +16,9 @@ from api.cas.cli_messages import (
     disallowed_param,
     operation_forbiden_in_standby,
     mutually_exclusive_params_init,
-    mutually_exclusive_params_load,
+    mutually_exclusive_params_load, activate_without_detach,
 )
-from api.cas.cache_config import CacheLineSize
+from api.cas.cache_config import CacheLineSize, CacheStatus
 from api.cas import cli
 from api.cas.ioclass_config import IoClass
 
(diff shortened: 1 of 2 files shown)


bkarpins:
retest this please

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -297,3 +297,48 @@ def test_start_neg_cli_flags():
...
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+def test_activate_without_detach():
+    """
+    title: Activate cache without detach command.
+    description: |
+       Try activate passive cache without detach command before activation.
+    pass_criteria:
+      - The activation is not possible
+      - The cache remains in Standby state after unsuccessful activation
+      - The cache exported object is present after an unsuccessful activation
```
mmichal10, line 311: The test doesn't verify if the cache exported object is present after an unsuccessful activation

pdebski21: Right, I moved check after cache activation.

mmichal10: Actually, I think there should be two checks - before and after the unsuccessful activation.

mmichal10: And one more thing; I know that the test is verifying the activate exit status and the error message but there should be an additional check to make sure that the cache is still running in standby mode

pdebski21: Good point. Now exported object presence is checked after load and unsuccessful activate.
After activation cache state is also checked.

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -297,3 +297,65 @@ def test_start_neg_cli_flags():
...
+
+    with TestRun.step("Verify if cache is in standby state after failed activation"):
+        caches = casadm_parser.get_caches()
+        if len(caches) < 1:
+            TestRun.LOGGER.error(f'Cache not present in system')
+        else:
+            cache_status = caches[0].get_status().value
+            if cache_status != "standby":
+                TestRun.LOGGER.error(
+                    f'Expected Cache state: "standby"'
```
mmichal10, line 354: You should use `CacheStatus.standby` enum (defined in `api/cas/cache_config.py`) instead of raw string.

mmichal10 (approved):

karolinavelkaja (approved):

PR #1203, opened 2022-05-11
Copyright header check improvements

rafalste:
Make sure to look for all files that need to have copyright header included.

Needs https://github.com/intel-innersource/frameworks.validation.opencas.test-framework/pull/61


rafalste:
retest this please

karolinavelkaja (dismissed):

configure.d/1_bd_part_count.conf
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-. $(dirname $3)/conf_framework
+. $(dirname $3)/conf_framework.sh
```
arutk, line 7: I think we should not update the copyright date in a files where the content doesn't change

arutk (changes requested):

configure.d/1_bd_part_count.conf
arutk: sorry, this comment is misplaced

tools/version2sha.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright(c) 2021 Intel Corporation
+# Copyright(c) 2021-2022 Intel Corporation
```
arutk, line 3: I think we should not update the copyright date in a files where the content doesn't change

rafalste: Fixed with this PR: https://github.com/Open-CAS/open-cas-linux/pull/1340

arutk (approved):

robertbaldyga (approved):

PR #1208, opened 2022-05-20
packaging: Fix debug package build

rafalste:
For debug symbols packages to build properly, we need to instruct the compiler to produce debugging information during the compilation process by adding a proper flag.
Additionally there is no point to create packages with debug info in normal build, because it may crash the package creation process if no debug info is found.


tools/pckgen
```diff
@@ -302,11 +302,25 @@ rpm_spec_prepare() {
...
+
+        if ! grep "$make_pattern" "$rpm_spec_file" &>/dev/null \
+                || ! grep "$debug_package_pattern" "$rpm_spec_file" &>/dev/null; then
+            error "couldn't find proper macros for debug build in file $rpm_spec_file"
+        fi
+        sed -i "/$make_pattern/ s/$/ DEBUG_PACKAGE=1/g" "$rpm_spec_file"
+        sed -i "/$debug_package_pattern/ s/%define debug_package %{nil}//g" "$rpm_spec_file"
         if [[ $(check_os) =~ suse|sles ]]; then
-            sed -i "s/%prep/%debug_package\n\n\n%prep/g" "$RPM_SPECS_DIR/$CAS_NAME.spec"
+            sed -i "s/%prep/%debug_package\n\n\n%prep/g" "$rpm_spec_file"
```
robertbaldyga, line 323: This series of `sed` commands greatly obfuscate way in which spec is generated without leaving a hint within the spec itself. I'd rather have template strings within spec file (like `<CAS_NAME>` or `<CAS_VERSION>`) that are populated by `pckgen` depending on its parameters. That way parts of the spec that are generated dynamically will be clearly separated from parts that are static.

tools/pckgen
```diff
@@ -349,6 +363,24 @@ deb_control_files_prepare() {
...
+        # debug symbols (to force debug package creation).
+        local make_pattern='make -C casadm$'
+        local strip_pattern='dh_strip.*--no-ddebs'
+
+        if ! grep "$make_pattern" "$deb_rules_file" &>/dev/null \
+                || ! grep "$strip_pattern" "$deb_rules_file" &>/dev/null; then
+            error "couldn't find proper rules for debug build in file $deb_rules_file"
+        fi
+        sed -i "/$make_pattern/ s/$/ DEBUG_PACKAGE=1/g" "$deb_rules_file"
+        sed -i "/$strip_pattern/ s/--no-ddebs/--ddebs/g" "$deb_rules_file"
```
robertbaldyga, line 382: The same comment here. Try to use templating instead.

robertbaldyga (changes requested):

tools/pckgen
rafalste: done

tools/pckgen
rafalste: done

tools/pckgen
```diff
@@ -298,6 +299,7 @@ rpm_spec_prepare() {
     sed -i "s/<CAS_NAME>/$CAS_NAME/g" "$RPM_SPECS_DIR/$CAS_NAME.spec"
     sed -i "s/<CAS_VERSION>/$CAS_VERSION/g" "$RPM_SPECS_DIR/$CAS_NAME.spec"
     sed -i "s/<CAS_LICENSE_NAME>/$CAS_LICENSE_NAME/g" "$RPM_SPECS_DIR/$CAS_NAME.spec"
+    sed -i "s|<CAS_MODULES_DIR>|$CAS_MODULES_DIR|g" "$RPM_SPECS_DIR/$CAS_NAME.spec"
```
robertbaldyga, line 302: Why `|`?

rafalste: Sed can take virtually any character as a separator. In this case we are replacing a directory path that already contains slashes, so it would interfere with sed separators if we'd used slashes here also.

robertbaldyga (approved):

jfckm (approved):

PR #1194, opened 2022-04-27
OpenCAS Linux kernel 5.15 adaptation

kmajzero:
This patch adapts OpenCAS Linux to compile and work with kernel v5.15

Signed-off-by: Krzysztof Majzerowicz-Jaszcz <krzysztof.majzerowicz-jaszcz@intel.com>

PR #1211, opened 2022-05-24
cas_disk: adapt to kernel 5.13

hammerg:
Kernel commit c76f48eb add a lock of bd_mutex in the del_gendisk
function. This cause a deadlock in cas_disk as it hold this lock
before calling del_gendisk.

This patch fix this issue by holding the bd_mutex only when
accessing the bd_openers field and releasing it afterward. It
also add an open ops function which prevents an access to the
disk while it is shutting down (locked_bd is set).

Signed-off-by: Gal Hammer <gal.hammer@huawei.com>
Signed-off-by: Shai Fultheim <shai.fultheim@huawei.com>

### DIFF
--- a/modules/cas_disk/exp_obj.c
+++ b/modules/cas_disk/exp_obj.c
@@ -1,5 +1,5 @@
 /*
-* Copyright(c) 2012-2021 Intel Corporation
+* Copyright(c) 2012-2022 Intel Corporation
 * SPDX-License-Identifier: BSD-3-Clause
 */
 #include <linux/module.h>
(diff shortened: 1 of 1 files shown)


jfckm:
Hi Gal,
Thanks for your contribution. We currently have a patch under development that will make CAS work with kernels up to 5.15 (#1194). I think it might address the issues you're trying to fix here. Do you mind checking it out?

hammerg:
> Hi Gal, Thanks for your contribution. We currently have a patch under development that will make CAS work with kernels up to 5.15 (#1194). I think it might address the issues you're trying to fix here. Do you mind checking it out?

No problem, I'll take a look.

hammerg:
Not required after the merge of PR #1194.

PR #1213, opened 2022-05-25
cas_cache: fix "blk_update_request: operation not supported" error

hammerg:
A write request with the REQ_RAHEAD flag enabled cause the nvme driver
to send a write command with access frequency value that is reserved
(at least on specification version 1.4c). NVME devices might fail this
write command with an unsupported error. So we now clear the request's
flag based on its direction.

Signed-off-by: Gal Hammer <gal.hammer@huawei.com>
Signed-off-by: Shai Fultheim <shai.fultheim@huawei.com>


jfckm:
@CAS-Linux-Jenkins ok to test

hammerg:
Anyone? Thanks, Gal.

rafalste (approved):

PR #1216, opened 2022-05-27
tests: Advanced failover test

mmichal10:


mmichal10:
retest this please

test/functional/tests/failover_standby/test_failover_multihost.py
```diff
@@ -0,0 +1,534 @@
...
+cls = CacheLineSize.LINE_32KiB
+
+
+@pytest.mark.require_disk("metadata_dev", DiskTypeSet([DiskType.nand]))
+@pytest.mark.require_disk("core_dev", DiskTypeSet([DiskType.hdd]))
+@pytest.mark.require_disk("raid_dev1", DiskTypeSet([DiskType.optane]))
+@pytest.mark.require_disk("raid_dev2", DiskTypeSet([DiskType.optane]))
+@pytest.mark.multidut(2)
+@pytest.mark.require_plugin("power_control")
+def test_functional_activate_twice_round_trip():
```
arutk, line 47: I suggest to parametrize two both tests to have a version with and without file system - this generates significantly different I/O pattern to CAS and potential debugging should be easier on a version without file system

test/functional/tests/failover_standby/test_failover_multihost.py
```diff
@@ -0,0 +1,534 @@
...
+    return caches_original_resource, caches_failover_resource, cores_resource
+
+
+def power_failure(primary_node, core_path):
+    with TestRun.use_dut(primary_node), TestRun.step("Make sure drbd instances are in sync"):
+        primary_node.cache_drbd.wait_for_sync()
+        primary_node.core_drbd.wait_for_sync()
+
+    with TestRun.use_dut(primary_node), TestRun.step("Get cache stats and core's md5"):
+        core_md5 = TestRun.executor.run(f"md5sum {core_path}").stdout.split()[0]
```
arutk, line 439: I think in order to make the test execution more predictable I would suggest to switch cache to WO mode for the time of checksum calculation (and switch back after calculation). Currently after failover md5 calculation results in multiple partial misses and the cache is cleaned as a result. This isn't bad, but it is unexpected.

This applies to this line as well as the other where md5 is calculated.

test/functional/tests/failover_standby/test_failover_multihost.py
```diff
@@ -0,0 +1,534 @@
...
+                cache_id=cache_id,
+                force=True,
+            )
+
+    with TestRun.use_dut(new_secondary_node), TestRun.step(
+        f"Start secondary DRBD on {new_secondary_node.ip}"
+    ):
+        drbd_resource.save()
+        new_secondary_node.cache_drbd = Drbd(drbd_resource)
+        new_secondary_node.cache_drbd_dev = new_secondary_node.cache_drbd.up()
```
arutk, line 534: In case of autoload == False (simulating new secondary node added to the cluster)  cache drbd resource must have metadata re-initialized on the new_secondary_node. I don't see it being done here.

arutk (changes requested):

mmichal10:
retest this please

arutk (approved):

PR #1236, opened 2022-06-22
Tests: Add test_standby_start_stop

KlaudiaJ:
Test if cache can be started in standby state and stopped without activation.

Signed-off-by: Klaudia Jablonska <klaudia.jablonska@intel.com>

### DIFF
--- a/test/functional/tests/cli/test_cli_standby.py
+++ b/test/functional/tests/cli/test_cli_standby.py
@@ -235,7 +235,7 @@ def test_standby_neg_cli_management():
             cli.flush_core_cmd(str(cache_id), "1"),
             cli.load_io_classes_cmd(str(cache_id), ioclass_config_path),
             cli.list_io_classes_cmd(str(cache_id), output_format="csv"),
-            cli.script_try_add_cmd(str(cache_id), core_device.path, core_id=1),
+            cli.script_try_add_cmd(str(cache_id), core_device.path, core_id="1"),
             cli.script_purge_cache_cmd(str(cache_id)),
             cli.script_purge_core_cmd(str(cache_id), "1"),
             cli.script_detach_core_cmd(str(cache_id), "1"),
@@ -414,10 +414,10 @@ def test_activate_neg_cache_line_size():
                 dd_count = int(md_size / Size(1, Unit.MebiByte)) + 1
                 (
                     Dd().input(active_cache_dev.path)
-                        .output(md_dump.full_path)
-                        .block_size(Size(1, Unit.MebiByte))
-                        .count(dd_count)
-                        .run()
+                    .output(md_dump.full_path)
+                    .block_size(Size(1, Unit.MebiByte))
+                    .count(dd_count)
+                    .run()
                 )
                 md_dump.refresh_item()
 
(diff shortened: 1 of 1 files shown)


mmichal10:
ok to test

kmajzero (approved):

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -459,3 +459,51 @@ def test_activate_neg_cache_line_size():
...
+            .block_size(block_size) \
+            .seek(offset)
+
+    with TestRun.step("Count and save a checksum for the written data on the exported object"):
+        cache_md5 = TestRun.executor.run(f"md5sum /dev/cas-cache-{cache_id}").stdout.split()[0]
+
+    with TestRun.step("Stop the cache instance"):
+        cache.stop()
+
+    with TestRun.step("Count anmd compare checksum from the cache and device"):
```
karolinavelkaja, line 506: typo *and

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -459,3 +459,52 @@ def test_activate_neg_cache_line_size():
...
+                      "object (with offset set to the middle of the device)"):
+        block_size = Size(4, Unit.MebiByte)
+        offset = int(cache_size/block_size/2)
+        Dd().output(f"/dev/cas-cache-{cache_id}") \
+            .input("/dev/urandom") \
+            .block_size(block_size) \
+            .seek(offset)
+
+    with TestRun.step("Count and save a checksum for the written data on the exported object"):
+        cache_md5 = TestRun.executor.run(f"md5sum /dev/cas-cache-{cache_id}").stdout.split()[0]
```
mmichal10, line 501: You should calculate md5 only for the part of cache exported object which you populated with data. The current implementation may fail because of the differences in metadata section after the standby instance is stopped

mmichal10:
Could you please squash commits (except the one with cleanup) into one?

test/functional/tests/cli/test_cli_standby.py
```diff
@@ -569,3 +569,51 @@ def test_standby_init_with_preexisting_filesystem(filesystem):
...
+
+    with TestRun.step("Count and save a checksum for the written data on the exported object"):
+        cache_md5 = TestRun.executor.run(f"dd if=/dev/cas-cache-{cache_id} skip={offset} | md5sum").stdout.split()[0]
+
+    with TestRun.step("Stop the cache instance"):
+        cache.stop()
+
+    with TestRun.step("Count and compare checksum from the cache and device"):
+        cache_device_md5 = TestRun.executor.run(f"dd if={cache_device.path} skip={offset} | md5sum").stdout.split()[0]
+        if cache_md5 != cache_device_md5:
```
arutk, line 618: Strictly speaking this doesn't check whether we read the same data that was originally written. This checks whether the same data is read from /dev/cas-cache-1 and the cache device. I think this should be changed so that we know exactly the checksum of the data that was written to /dev/cas-cache-1.

arutk (changes requested):

arutk:
looks good to me. one think I would think about is removing I/O offset (seek/skip) so that entire device is being written and/or read.

mmichal10 (approved):

mmichal10 (approved):

arutk (approved):

PR #1265, opened 2022-07-05
Fix core pool tests 

pdebski21:


mmichal10 (dismissed):

arutk (approved):

test/functional/tests/incremental_load/test_core_pool.py
```diff
@@ -48,10 +48,11 @@ def test_attach_core_pool():
         cache.stop()
```
karolinavelkaja, line 49: Please update the date (year) in file

test/functional/tests/incremental_load/test_core_pool.py
```diff
@@ -48,10 +48,11 @@ def test_attach_core_pool():
...
 
     with TestRun.step("Add previously used core device to core pool using --try-add flag."):
-        first_core = casadm.try_add(core_dev, cache.cache_id)
+        first_core = casadm.try_add(core_device=core_dev, cache_id=cache.cache_id, core_id=1)
```
karolinavelkaja, line 51: Instead of magic numbers "1", and "2", could you use "first_core.core_id" and "second_core.core_id"if available please? Or define for example "core_1_id =1" and "core_2_id =2" and use them here and later instead of numbers

pdebski21: I defined additional variables for core and cache id.

karolinavelkaja (approved):

test/functional/tests/incremental_load/test_core_pool.py
karolinavelkaja: thank you  🤩🌼

robertbaldyga (approved):

PR #1267, opened 2022-07-05
API fix for casadm standby init

kmajzero:
Cache line size parameter is handled inconsistently in standby init and
start cache methods.

This patch fixes this inconsistency so both methods accept a proper
CacheLineSize object.

Some of the existing tests needed fixing as well, as included in this
patch.

### DIFF
--- a/test/functional/api/cas/casadm.py
+++ b/test/functional/api/cas/casadm.py
@@ -47,11 +47,15 @@ def standby_init(cache_dev: Device, cache_id: int, cache_line_size: CacheLineSiz
                  kernel_params: KernelParameters = KernelParameters()):
     if kernel_params != KernelParameters.read_current_settings():
         reload_kernel_module("cas_cache", kernel_params.get_parameter_dictionary())
+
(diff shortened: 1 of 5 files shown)


mmichal10 (approved):

test/functional/api/cas/casadm.py
```diff
@@ -47,11 +47,15 @@ def standby_init(cache_dev: Device, cache_id: int, cache_line_size: CacheLineSiz
                  kernel_params: KernelParameters = KernelParameters()):
     if kernel_params != KernelParameters.read_current_settings():
         reload_kernel_module("cas_cache", kernel_params.get_parameter_dictionary())
+
+    _cache_line_size = None if cache_line_size is None else str(
+        int(cache_line_size.value.get_value(Unit.KibiByte)))
```
robertbaldyga, line 52: `get_value()` returns `int` so there is no need to cast.

robertbaldyga (approved):

PR #1288, opened 2022-07-22
Couple of improvements for faster tests

jfckm:
Needs changes in test-framework. Also, fails sporadically while removing partitions - still needs some debugging

### DIFF
--- a/test/functional/tests/conftest.py
+++ b/test/functional/tests/conftest.py
@@ -259,10 +259,7 @@ def base_prepare(item):
         for disk in TestRun.disks.values():
             disk_serial = get_disk_serial_number(disk.path)
             if disk.serial_number != disk_serial:
-                raise Exception(
-                    f"Serial for {disk.path} doesn't match the one from the config."
-                    f"Serial from config {disk.serial_number}, actual serial {disk_serial}"
-                )
+                disk.serial_number = disk_serial
 
             disk.umount_all_partitions()
             Mdadm.zero_superblock(os.path.join('/dev', disk.device_id))
--- a/test/functional/tests/conftest.py
+++ b/test/functional/tests/conftest.py
@@ -256,7 +256,7 @@ def base_prepare(item):
 
         RamDisk.remove_all()
 
-        for disk in TestRun.dut.disks:
+        for disk in TestRun.disks.values():
             disk_serial = get_disk_serial_number(disk.path)
             if disk.serial_number != disk_serial:
                 raise Exception(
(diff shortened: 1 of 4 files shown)


test/functional/tests/conftest.py
```diff
@@ -253,24 +253,21 @@ def base_prepare(item):
...
         RamDisk.remove_all()
 
-        for disk in TestRun.dut.disks:
+        for disk in TestRun.disks.values():
             disk_serial = get_disk_serial_number(disk.path)
             if disk.serial_number != disk_serial:
-                raise Exception(
-                    f"Serial for {disk.path} doesn't match the one from the config."
-                    f"Serial from config {disk.serial_number}, actual serial {disk_serial}"
-                )
```
pdebski21, line 266: After this change is disk serial still needed in dut config file?
Don't we want to have some type of info that serial from config does not match?

karolinavelkaja (approved):

mmichal10 (approved):

test/functional/tests/conftest.py
jfckm: It's not needed, but for compatibility's sake I left it in (no need to go and fix all your configs). IMO you shouldn't have to fill it if tests get it anyway. As for info that serials are different - what value do you expect from this info?

pdebski21 (dismissed):

kmajzero (commented):
What if a particular test would need custom setup/teardown? Is this possible (or even needed)?

jfckm:
> What if a particular test would need custom setup/teardown? Is this possible (or even needed)?

Right now I believe it's handled by doing prepare/teardown inline in test or separating it out to per-module functions. You could also use [fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html), but (sadly) they don't get much use in our TF/tests.

kmajzero (approved):

test/functional/tests/conftest.py
```diff
@@ -57,61 +51,53 @@ def pytest_runtest_setup(item):
...
+        TestRun.LOGGER.info(f"CAS cleanup on {dut.ip}")
+        remove(opencas_drop_in_directory, recursive=True, ignore_errors=True)
 
-    test_name = item.name.split('[')[0]
-    TestRun.LOGGER = create_log(item.config.getoption('--log-path'), test_name)
+        try:
+            InitConfig.create_default_init_config()
+            unmount_cas_devices()
+            try:
+                casadm.stop_all_caches()
```
arutk, line 65: isn't DRBD missing from here?

jfckm: it is - I moved it together with all other prepare stuff to testframework https://github.com/intel-innersource/frameworks.validation.opencas.test-framework/pull/52

jfckm:

Deixx:
needs serious analysis, preferably by the author  
1. conflicts (not surprising after this time)
2. test-framework changes implied (is there a corresponding MR for TF?)

PR #1289, opened 2022-07-26
tests: fix step "Try to zero metadata with 'sudo'." in test_user_cli

karolinavelkaja:
fix: zeroing metadata on running cache expects to fail.

### DIFF
--- a/test/functional/tests/security/test_security_user.py
+++ b/test/functional/tests/security/test_security_user.py
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2019-2021 Intel Corporation
+# Copyright(c) 2019-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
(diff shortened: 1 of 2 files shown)


pdebski21 (approved):

pdebski21:
It would be better to move file formatting with black to another commit.

test/functional/tests/security/test_security_user.py
```diff
@@ -354,10 +354,12 @@ def test_user_cli():
 
     with TestRun.step("Try to zero metadata with 'sudo'."):
         try:
-            run_as_other_user(cli.zero_metadata_cmd(str(cache_dev)),
-                              user_name, True)
+            output = run_as_other_user(cli.zero_metadata_cmd(str(cache_dev.path)), user_name, True)
+            if output.exit_code == 0:
```
robertbaldyga, line 358: This `if` is actually unnecessary. We expect `run_as_other_user()` to raise an exception, so if that does not happen we can just error out.

karolinavelkaja: I may not understand ^^ this "if statement " and next line is for the error case = " if that does not happen we can just error out".
This is exactly what you've proposed I guess. And I think (and hope) the messages are helpful for understanding tests intentions and what is expected here.

pdebski21 (approved):

jfckm (approved):

PR #1293, opened 2022-07-27
tests: fix test_cli_help

karolinavelkaja:

### DIFF
--- a/test/functional/tests/cli/test_cli_help_and_version.py
+++ b/test/functional/tests/cli/test_cli_help_and_version.py
@@ -1,10 +1,11 @@
 #
-# Copyright(c) 2020-2021 Intel Corporation
+# Copyright(c) 2020-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-import pytest
 import re
+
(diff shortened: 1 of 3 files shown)


pdebski21 (approved):

test/functional/tests/cli/test_cli_help_and_version.py
```diff
@@ -1,11 +1,12 @@
...
-# Copyright(c) 2020-2021 Intel Corporation
+# Copyright(c) 2020-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-import pytest
 import re
 
+import pytest
+
```
robertbaldyga, line 9: Why this change? Most of the tests import pytest before other imports. Do we want to change this convention?

karolinavelkaja: I suspect this is result of "optimize imports", I don't recall to do such changes, there was no need. I can revert this if it can make any difference - does it?

mmichal10 (approved):

pdebski21 (approved):

PR #1295, opened 2022-08-02
Test cleaning policy.

pskwarlo:
Signed-off-by: Pawel Skwarlo <pawel.skwarlo@intel.com>

### DIFF
--- a/test/functional/tests/cli/test_cleaning_policy.py
+++ b/test/functional/tests/cli/test_cleaning_policy.py
@@ -1,0 +1,77 @@
+#
+# Copyright(c) 2022 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+import pytest
+import time
+
+from core.test_run_utils import TestRun
+from test_utils.size import Size, Unit
+from core.test_run import TestRun
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from test_tools.fio.fio import Fio
+from test_tools.fio.fio_param import ReadWrite, IoEngine
+from api.cas import casadm
+from api.cas.cache_config import CacheMode, CleaningPolicy
+from test_utils.os_utils import Udev
+
+wait_time_s = 30
+
+
+@pytest.mark.CI
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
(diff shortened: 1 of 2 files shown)


test/functional/tests/cli/test_cleaning_policy.py
```diff
@@ -0,0 +1,72 @@
...
+    Title: test_cleaning_policy
+    description: | The test is to see if dirty data will be removed from the Cache
+    after changing the cleaning policy from NOP to one that expects a flush.
+    pass_criteria:
+        - Cache is successfully populated with dirty data
+        - Cleaning policy is changed successfully
+        - There is no dirty data after the policy change
+    """
+    with TestRun.step("Prepare devices."):
+        wait_time_s = 2
```
karolinavelkaja, line 35: this could be placed above pytest marks :)

karolinavelkaja (dismissed):

test/functional/tests/cli/test_cleaning_policy.py
```diff
@@ -0,0 +1,81 @@
...
+
+    with TestRun.step(f"Change cleaning policy and wait up to {wait_time_s} for flush"):
+        cache.set_cleaning_policy(CleaningPolicy.acp)
+        t_end = time.time() + wait_time_s
+        while time.time() < t_end:
+            if cache.get_dirty_blocks() == Size.zero():
+                TestRun.LOGGER.info(
+                    f"Cache flushed after {round(time.time() - (t_end - wait_time_s), 2)} seconds"
+                )
+                break
```
robertbaldyga, line 77: I'd add some short sleep inside this loop, lets say 500ms or 1s. It doesn't make sense to check stats more often than that. Moreover, checking stats very frequently can slow down cleaning (or even block it entirely), as cleaner thread attempts to take mngt lock.

robertbaldyga (changes requested):

test/functional/tests/cli/test_cleaning_policy.py
pskwarlo: I added a sleep method depending on the total time we set to wait, for 1 minute wait time inside loop is 1 second.

karolinavelkaja (approved):

robertbaldyga (approved):

PR #1296, opened 2022-08-08
Recovery tests - CI

pdebski21:
Bash tests rewritten to test-framework form open-cas-linux/test/smoke_test/recovery

### DIFF
--- a/test/functional/tests/ci/test_recovery.py
+++ b/test/functional/tests/ci/test_recovery.py
@@ -1,0 +1,210 @@
+#
+# Copyright(c) 2019-2022 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+import pytest
+
+from api.cas import casadm
+from api.cas.cache_config import CacheMode
+from api.cas.cli import casadm_bin
+from api.cas.cli_messages import check_stderr_msg, stop_cache_errors
+from core.test_run import TestRun
+from storage_devices.disk import DiskTypeLowerThan, DiskTypeSet, DiskType, Disk
+from test_tools.dd import Dd
+from test_tools.disk_utils import Filesystem, unmount, mount
+from test_tools.fs_utils import check_if_file_exists
+from test_utils.filesystem.file import File
+from test_utils.os_utils import sync
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/cas"
+
+
+@pytest.mark.CI
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_recover_cache_verify_core():
+    """
+        title: Recovery after turning off/on devices
+        description: |
+            Test data integrity after turning off cache device.
+        pass_criteria:
+          - Cache devices successfully loaded with metadata after turning devices off/on
+          - md5sums before and after all operations match each other
+          - creation, mount, unmount of filesystems on the core device succeeds
+    """
+    filesystems = [Filesystem.xfs, Filesystem.ext3, Filesystem.ext4]
+    cache_cnt = len(filesystems)
+
+    with TestRun.step("Prepare devices."):
+        cache_disk = TestRun.disks["cache"]
+        cache_disk.create_partitions([Size(2, Unit.GibiByte)] * cache_cnt)
+        cache_devs = cache_disk.partitions
+        core_disk = TestRun.disks["core"]
+        core_disk.create_partitions([Size(4, Unit.GibiByte)] * cache_cnt)
+        core_devs = core_disk.partitions
+
+    with TestRun.step("Start caches and add cores."):
+        caches, cores = [], []
+        for (cache_dev, core_dev) in zip(cache_devs, core_devs):
+            cache = casadm.start_cache(cache_dev, cache_mode=CacheMode.WB)
+            core = cache.add_core(core_dev)
+            caches.append(cache)
+            cores.append(core)
+
+    with TestRun.step("Create filesystem on core devices."):
+        for (core, filesystem) in zip(cores, filesystems):
+            core.create_filesystem(filesystem)
+
+    with TestRun.step("Mount cache devices."):
+        for (cache, core) in zip(caches, cores):
+            core_mnt_point = f"{mount_point}-{cache.cache_id}-{core.core_id}"
+            core.mount(core_mnt_point)
+
+            with TestRun.step("Run IO"):
+                dd = (
+                    Dd()
+                    .input("/dev/urandom")
+                    .output(f"{core_mnt_point}/test")
+                    .count(1)
+                    .block_size(Size(50, Unit.MegaByte))
+                )
+                dd.run()
+
+    with TestRun.step("Calculate cache md5sums before unplug."):
+        core_mnt_md5s_before = [File(f"{core.mount_point}/test").md5sum() for core in cores]
+
+    with TestRun.step("Umount core devices."):
+        for core in cores:
+            core.unmount()
+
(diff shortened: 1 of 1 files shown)


test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+    with TestRun.step("Start caches and add cores."):
+        caches, cores = [], []
+        for (cache_dev, core_dev) in zip(cache_devs, core_devs):
+            cache = casadm.start_cache(cache_dev, cache_mode=CacheMode.WB)
+            core = cache.add_core(core_dev)
+            caches.append(cache)
+            cores.append(core)
+
+    with TestRun.step("Create filesystem on core devices."):
+        filesystems = [Filesystem.xfs, Filesystem.ext3, Filesystem.ext4]
```
arutk, line 54: for easier future modification you might consider moving this line to the beginning and then replacing "3" with named variable, e.g. "num_caches = len(filesystems)"

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+        for (core, filesystem) in zip(cores, filesystems):
+            core.create_filesystem(filesystem)
+
+    with TestRun.step("Mount cache devices."):
+        core_mount_points = []
```
arutk, line 59: I don't think you need to maintain core_mount_points, this is just an equivalent of [c.mount_point for c in cores[

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+            with TestRun.step("Run IO"):
+                dd = (
+                    Dd()
+                    .input("/dev/urandom")
+                    .output(f"{core_mnt_point}/test")
+                    .count(1)
+                    .block_size(Size(50, Unit.MegaByte))
+                )
+                dd.run()
+                sync()
```
arutk, line 74: unnecessary, you can remove this

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def dirty_stop(cache_disk, caches: list):
+    with TestRun.step("Turn off cache devices."):
+        cache_disk.unplug()
+
+    with TestRun.step("Stop caches without flushing."):
+        for cache in caches:
+            with pytest.raises(CmdException):
+                cache.stop(no_data_flush=True)
```
arutk, line 224: maybe verify the right type of exception is catched?

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+    with TestRun.step("Compare md5 sums for cores and core devices"):
+        if core_mnt_md5s_before != core_mnt_md5s_after:
+            TestRun.fail(f"MD5 sums of core before and after does not match.")
+
+    with TestRun.step("Umount core devices."):
+        for core_dev in core_devs:
+            unmount(core_dev)
+
+    with TestRun.step("Dirty stop"):
+        dirty_stop(cache_disk, caches)
```
arutk, line 116: was it there in the original test? seems to be unnecessary

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+        ]
+
+    with TestRun.step("Umount core devices."):
+        for core in cores:
+            core.unmount()
+
+    with TestRun.step("Dirty stop"):
+        dirty_stop(cache_disk, caches)
+
+    with TestRun.step("Start caches with load metadata and later stop them."):
```
arutk, line 183: no stop here in this step

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_recover_cache():
+    """
+        title: Recovery after turning off/on devices
+        description: |
+            Test data integrity after turning off cache device.
+        pass_criteria:
+          - Cache devices successfully loaded with metadata after turning devices off/on
+          - md5sums before and after all operations match each other
+          - creation, mount, unmount of filesystems succeeds
```
arutk, line 35: I would add "... on the core device"

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,227 @@
...
+from test_utils.output import CmdException
+from test_utils.size import Size, Unit
+
+mount_point = "/mnt/cas"
+
+
+@pytest.mark.CI
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.optane, DiskType.nand]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+def test_recover_cache():
```
arutk, line 27: I would rename the tests to better reflect the difference:
this one: test_recover_cache_verify_core()
the next one: test_recover_cache_verify_exp_obj()

arutk (changes requested):

test/functional/tests/ci/test_recovery.py
pdebski21: Yes it was but prepare and clean up form test framework may be enough.

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,218 @@
...
+                TestRun.LOGGER.error(f"Mounting core device {core_mnt_point} failed.")
+
+    with TestRun.step("Calculate cache md5sums after recovery."):
+        core_mnt_md5s_after = [
+            File(f"{core.mount_point}/test").md5sum() for core in cores
+        ]
+
+    with TestRun.step("Compare md5 sums for cores and core devices"):
+        if core_mnt_md5s_before != core_mnt_md5s_after:
+            TestRun.fail(f"MD5 sums of core before and after does not match.")
```
mmichal10, line 106: If md5s don't match, I think it would be valuable to print both lists to allow seeing which one is invalid.

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,218 @@
...
```
mmichal10, line 193: If md5s don't match, I think it would be valuable to print both lists to allow seeing which one is invalid.

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,218 @@
...
+    with TestRun.step("Umount core devices."):
+        for core in cores:
+            core.unmount()
+
+    with TestRun.step("Stop caches"):
+        for cache in caches:
+            cache.stop()
+
+    with TestRun.step("Dirty stop"):
+        dirty_stop(cache_disk, caches)
```
mmichal10, line 204: You've just stopped all running caches so I think this step seem redundant

pdebski21: I have removed stop steps as in previous test.

test/functional/tests/ci/test_recovery.py
```diff
@@ -0,0 +1,209 @@
...
+
+def dirty_stop(cache_disk, caches: list):
+    with TestRun.step("Turn off cache devices."):
+        cache_disk.unplug()
+
+    with TestRun.step("Stop caches without flushing."):
+        for cache in caches:
+            with pytest.raises(CmdException) as ex:
+                cache.stop(no_data_flush=True)
+            assert "Failed to stop cache." in str(ex.value)
```
mmichal10, line 206: You should try to keep off `asserts` and use `TestRun.fail()` instead. Or even better, you can try with `api/cas/cli_messages.py` api.

mmichal10 (approved):

arutk (approved):

PR #1307, opened 2022-08-12
Integrate kmemleak into CAS

mmichal10:
Kmemleak is a tool which allows to detect memory leaks in kernel. It requires
`CONFIG_DEBUG_KMEMLEAK` to be enabled during kernel compilation. Oterwise all
the kmemleak_* functions are NOPs.

### DIFF
--- a/modules/cas_cache/context.c
+++ b/modules/cas_cache/context.c
@@ -1,5 +1,5 @@
 /*
-* Copyright(c) 2012-2021 Intel Corporation
+* Copyright(c) 2012-2022 Intel Corporation
 * SPDX-License-Identifier: BSD-3-Clause
 */
 
@@ -10,6 +10,7 @@
 #include "utils/utils_gc.h"
 #include "utils/utils_mpool.h"
 #include "threads.h"
+#include <linux/kmemleak.h>
 
 struct env_mpool *cas_bvec_pool;
 
@@ -66,6 +67,7 @@ void *_cas_alloc_page_rpool(void *allocator_ctx, int cpu)
 
 	_cas_page_set_priv(page);
 	_cas_page_set_cpu(page, cpu);
+
 	return page_address(page);
 }
 
@@ -103,6 +105,11 @@ ctx_data_t *__cas_ctx_data_alloc(uint32_t pages, bool zalloc)
 			_cas_page_set_cpu(data->vec[i].bv_page, cpu);
 		} else {
 			data->vec[i].bv_page = alloc_page(GFP_NOIO);
+			if (data->vec[i].bv_page) {
+				/* Failed to get memory from rpool but backup allocation worked.
+				   Need to keep track of this page as well */
+				kmemleak_alloc(page_address(data->vec[i].bv_page), PAGE_SIZE, 1, GFP_NOIO);
+			}
 		}
 
 		if (!data->vec[i].bv_page)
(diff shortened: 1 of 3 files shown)


jfckm (approved):

pdebski21 (approved):

modules/cas_cache/context.c
```diff
@@ -66,6 +67,11 @@ void *_cas_alloc_page_rpool(void *allocator_ctx, int cpu)
 
 	_cas_page_set_priv(page);
 	_cas_page_set_cpu(page, cpu);
+
+	/* As it isn't an actuall allocation but only a rpool preparation
+	   kmemleak shouldn't track the allocated memory yet. */
+	kmemleak_free(page_address(page));
+
```
arutk, line 74: I think this is not correct, since you mentioned page allocations are not tracked by kmemleak

modules/cas_cache/ocf_env.c
```diff
@@ -194,6 +204,8 @@ void env_allocator_del(env_allocator *allocator, void *obj)
 			return;
 	}
 
+	/* It wasn't a page from rpool thus need to stop tracking it explicitly */
+	kmemleak_free(obj);
 	kmem_cache_free(allocator->kmem_cache, item);
```
arutk, line 209: shouldn't you do kmemleak_alloc(item) before kmem_cache_free so that this is not detected as double free?

modules/cas_cache/ocf_env.c
```diff
@@ -54,6 +55,12 @@ void *env_allocator_new(env_allocator *allocator)
 		BUG_ON(item->used);
 	} else {
 		item = kmem_cache_zalloc(allocator->kmem_cache, GFP_NOIO);
+		if (item) {
+			/* Kmemleak is supposed to track the memory CAS is actually using,
+			   not the container */
+			kmemleak_free(item);
+			kmemleak_alloc(&item->data, allocator->item_size, 1, GFP_NOIO);
```
arutk, line 62: Is this actually necessary? According to kmemleak documentation this will not be detected as a leak as long as &item->data is not lost.

"An allocated block of memory is considered orphan if no pointer to its start address or to any location inside the block can be found by scanning the memory"

arutk (changes requested):

modules/cas_cache/context.c
robertbaldyga: +1, normally it will not be very visible, but it will generate warning on debug kernel.

modules/cas_cache/ocf_env.c
robertbaldyga: Won't we get leak report if the scan occurs just between line 62 and line 70? There is no location in memory that holds pointer to `&item->data` in this window.

robertbaldyga (changes requested):

modules/cas_cache/ocf_env.c
arutk: @robertbaldyga In general I think it is possible that some pointers are temporarily stored in registers (or being moved between memory/registers etc) and kmemleak scanning at the wrong moment would identify these allocations as leaked. Making sure that allocation is referenced in memory at all times might not be worth the effort. Instead, since scan is ran on demand, it would be a good practice to run it when pointers are not being passed around.

robertbaldyga (approved):

kmajzero (approved):

PR #1308, opened 2022-08-12
Rework for RHEL8.5

kmajzero:
Rework for RHEL8.5 compatibility to avoid potential bug of exiting queue
after IO completion on a different CPU.

Targeted RHEL kernel version corrected.

### DIFF
--- a/configure.d/1_blk_queue_exit_workaround_for_rhel.conf
+++ b/configure.d/1_blk_queue_exit_workaround_for_rhel.conf
@@ -6,14 +6,14 @@
 
 . $(dirname $3)/conf_framework
 
-# RHEL8.5 kernel 4.18 subversion targeted for the workaround is 348.2.1.
+# RHEL8.5 kernel 4.18 subversion targeted for the workaround is 348.7.1.
 # The workaround is needed starting this version and up.
-LINUX_MAJOR_SINCE=4
-LINUX_MINOR_SINCE=18
-LINUX_SUB_SINCE=0
-RHEL_MAJOR_SINCE=348
-RHEL_MINOR_SINCE=2
-RHEL_SUB_SINCE=1
+QUEUE_PUT_WA_LINUX_MAJOR_SINCE=4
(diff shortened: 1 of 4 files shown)


jfckm (dismissed):

mmichal10 (approved):

configure.d/1_blk_queue_exit_for_rhel.conf
```diff
@@ -12,7 +12,7 @@ LINUX_MAJOR_SINCE=4
 LINUX_MINOR_SINCE=18
 LINUX_SUB_SINCE=0
 RHEL_MAJOR_SINCE=348
-RHEL_MINOR_SINCE=2
```
arutk, line 15: maybe prefix these variables names with "QUEUE_PUT_WA_" or something similar to indicate what this is about

configure.d/1_blk_queue_exit_for_rhel.conf
```diff
@@ -12,7 +12,7 @@ LINUX_MAJOR_SINCE=4
...
+RHEL_MINOR_SINCE=7
```
arutk, line 15: Have you considered limiting the scope of workaround up to some RHEL version? I'm not sure what would be the right approach. If this gets fixed at some point in a newer version of RHEL kernel 4.18.0 then CAS will start failing due to incorrect application of the workaround. 

Maybe we should enable workaround only on this specific kernel version and fail compilation on any newer RHEL 4.18.0 kernel? We'll still be working with newer kernels via weak modules mechanism, so there will be some limited room for safe upgrade.

arutk (changes requested):

configure.d/1_blk_queue_exit_for_rhel.conf
kmajzero: As discussed, we should leave it as it is and file a bug to RH. We would then specify the lowest and highest version numbers when (or if at all) it's fixed.

mmichal10 (approved):

arutk (approved):

arutk (approved):

PR #1311, opened 2022-08-16
Install particular CAS version on DUT before test

rafalste:
Check dut_config.yml for a CAS version to install. If none was specified, take local repo's HEAD as version to install.

### DIFF
--- a/test/functional/api/cas/git.py
+++ b/test/functional/api/cas/git.py
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2019-2021 Intel Corporation
+# Copyright(c) 2019-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
@@ -7,6 +7,7 @@ import os
 
 from core.test_run import TestRun
 from connection.local_executor import LocalExecutor
+from test_utils.output import CmdException
 
 
 def get_current_commit_hash(from_dut: bool = False):
@@ -25,6 +26,21 @@ def get_current_commit_message():
         f'git show HEAD -s --pretty=format:"%B"').stdout
 
 
+def get_commit_hash(cas_version, from_dut: bool = False):
+    executor = TestRun.executor if from_dut else LocalExecutor()
+    repo_path = TestRun.usr.working_dir if from_dut else TestRun.usr.repo_dir
+
+    output = executor.run(
+        f"cd {repo_path} && "
+        f"git rev-parse {cas_version}")
+    if output.exit_code != 0:
+        raise CmdException(f"Failed to resolve '{cas_version}' to commit hash", output)
+
+    TestRun.LOGGER.info(f"Resolved '{cas_version}' as commit {output.stdout}")
+
+    return output.stdout
+
+
 def get_release_tags():
     repo_path = os.path.join(TestRun.usr.working_dir, ".git")
     output = TestRun.executor.run_expect_success(f"git --git-dir={repo_path} tag").stdout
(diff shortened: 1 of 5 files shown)


karolinavelkaja (approved):

test/functional/config/example_dut_config.yml
```diff
@@ -14,6 +14,15 @@ type: "local"
...
 working_dir: "/tmp/open-cas-linux/"
 
+# CAS version to test.
+# This version will be installed on DUT before test. Can be any string
+# digestible by git-checkout. If not specified, current state of local
+# repo (HEAD) will be used as version to install.
+# NOTE: Make sure to have a local master branch synced with upstream
+# as the local repo will be taken for version reference. To do this
+# simply run: `git fetch origin master:master`
+cas_version: "master"
```
mmichal10, line 24: What if I don't need the newest OCL version on my DUT and I want to avoid reinstallation? Is is possible to skip it without putting the current commit version into dut config file? If not, I think there should be such an option available.

rafalste: I assumed that you always want to run a test on a particular version of CAS, will it be the newest master, a version tag or just a version that you're checked out on your local repo from which you run that test. If this version is already installed, there is no reinstallation.
This patch eliminates the risk of using a random CAS version installed on DUT in the past.

Is there a scenario in which using such random version would be beneficial?

rafalste: Or maybe your question is about using your "current commit version" of CAS? If so, you just need to remove or comment out the 'cas_version' in your dut_config and then your current commit from your local repo will be taken as version to check (and install if not already).

test/functional/api/cas/git.py
```diff
@@ -36,29 +52,17 @@ def get_release_tags():
...
         f"cd {TestRun.usr.working_dir} && "
         f"git checkout --force {commit_hash}")
     if output.exit_code != 0:
-        raise CmdException(f"Failed to checkout to CAS {commit_hash}", output)
+        raise CmdException(f"Failed to checkout to {commit_hash}", output)
 
     output = TestRun.executor.run(
         f"cd {TestRun.usr.working_dir} && "
-        f"git submodule update")
+        f"git submodule update --force")
```
jfckm, line 66: This `--force` makes me a bit uncomfortable with `LocalExecutor`... I'm not sure I want tests to go and overwrite my repo state - I'd rather it errored out and let me know it failed

jfckm (changes requested):

test/functional/tests/conftest.py
```diff
@@ -275,12 +275,14 @@ def base_prepare(item):
             disk.remove_partitions()
             create_partition_table(disk, PartitionTable.gpt)
 
+        cas_version = TestRun.config.get("cas_version") or git.get_current_commit_hash()
```
jfckm, line 278: IMO it would be more suited for a command line parameter - the same as `--force-reinstall`. Something like `pytest --dut-config=config.yml --cas-version <master,hash,current> (...)` - I'm thinking current would serve the purpose @mmichal10 mentioned - you use the current repo state and that's it

jfckm (changes requested):

test/functional/api/cas/git.py
rafalste: All the executor operations are performed on the repo which resides in the 'working_dir', which is '/tmp/open-cas-linux'. Repo which resides there is constantly a subject to various actions - it is removed, changed, rsynced, etc. by many steps performed by test-framework.
So forcing the submodule update there is not the issue. It is not changing your local repo, just a copy of it.
Besides, if you look 6 lines above, there is a similar '--force' flag when checking out to a different revision (and this would be more dangerous, as it discards your unsaved changes). But again - this is only a working copy of the repo. :)

jfckm: Oh, sorry, didn't catch that, carry on!

test/functional/tests/conftest.py
rafalste: I'm not sure if moving the version specification from dut_config to command line parameter would be beneficial, because we want to have the ability to select a CAS version per DUT in the multidut scenario.
The other thing about "current" version - you just need to remove/comment-out the 'cas_version' in dut_config to use your current repo state as the version to check (and install if not already).

jfckm: OK, if default is to use the current repo state then it works for me. It's true about multidut - I don't think I'll ever use it that way, but maybe this flexibility will come in handy sometime in the future.

jfckm (dismissed):

jfckm (approved):

mmichal10 (approved):

PR #1315, opened 2022-08-18
Add test test_cli_help_spelling

AdamSeyda:


mmichal10:
ok to test

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,230 @@
...
+
+def test_cli_help_spelling():
+    """
+    title: Spelling test for 'help' command
+    description: Validates spelling of 'help' in CLI
+    pass criteria:
+    - no spelling mistakes are found
+    """
+
+    personal_directory = "/root/open-cas-linux/test/smoke_test/cas_ex.en.pws"
```
mmichal10, line 24: I'm not sure whether we want to rely on an existing config file, especially if it is located somewhere in repo. I suggest converting this config file into a list within a test and spawning it on a DUT during each test execution.

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,230 @@
...
+        second = "{ cmd=0 }"
+        third = "{ print $0 }"
+        backslash = "\\"
+
+        commands = TestRun.executor.run_expect_success(
+            f"{casadm_bin} -H | awk '/Available commands:/{first} /For detailed help/ {second} "
+            f"cmd {third}' | grep -o '{backslash}-{backslash}-{backslash}S*'")
+
+        commands = str(commands).split()
+        commands = commands[3:-1]
```
rafalste, line 45: I'm not sure if string representation of output from run_expect_success() will equal the stdout part of that output (which I assume was your goal). And also you probably wanted to use splitlines() rather than split() (which by default splits strings by spaces). I would put it more like this: `commands = output.stdout.splitlines()[3:-1]`

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,230 @@
...
+
+        output = TestRun.executor.run_expect_success(
+            f"{casadm_bin} {commands} -H | aspell list --lang=en_US "
+            f"--add-extra-dicts={personal_directory}")
+        if output.stdout:
+            TestRun.LOGGER.error("Misspelled words found:\n")
+            TestRun.LOGGER.error(output.stdout)
+
+    return 0
```
rafalste, line 54: No need for an empty return in Python. It is implicitly assumed when absent. :)

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,230 @@
...
+
+        commands = str(commands).split()
+        commands = commands[3:-1]
+
+        output = TestRun.executor.run_expect_success(
+            f"{casadm_bin} {commands} -H | aspell list --lang=en_US "
+            f"--add-extra-dicts={personal_directory}")
+        if output.stdout:
+            TestRun.LOGGER.error("Misspelled words found:\n")
+            TestRun.LOGGER.error(output.stdout)
```
rafalste, line 52: I believe you wanted to run this inside `for command in commands:`

test/functional/tests/cli/test_cli_help_spelling.py
rafalste: Moreover, we can not assume the repo location, because users can put their repo in any directory.

test/functional/tests/cli/test_cli_help_spelling.py
rafalste: And also - why excluding first three and the last command from spell check?

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,230 @@
...
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.CI()
+def test_cli_write_mode():
+    """
+    title: Verification test for write mode: read only
+    description: Verify if write mode: read only, works as expected and cache only reads
+    and do not cache writes
+    pass criteria:
+    - cache does not cache write
+    """
```
rafalste, line 67: This test doesn't check spelling. It should probably be put somewhere in "ci" or "cache_ops" folder.

Kgierszx:
retest this please

Deixx:
retest this please

test/functional/tests/ci/test_ci_write_mode_read_only.py
```diff
@@ -0,0 +1,187 @@
...
+            tmp_cache = tmp_cache[3]
+            read_cache_1 = TestRun.executor.run_expect_success(
+                f"iostat {cache_device.parent_device.path} | grep {tmp_cache} | "
+                f"awk 'NR==1 {temp}'").stdout
+            TestRun.executor.run_expect_success(
+                f"dd if=/dev/cas1-1 of=$TMP_DIR/file001.bin bs=4096 count={count} "
+                f"skip=10000 iflag=direct")
+            read_cache_2 = TestRun.executor.run_expect_success(
+                f"iostat {cache_device.parent_device.path} | grep {tmp_cache} | "
+                f"awk 'NR==1 {temp}'").stdout
```
Deixx, line 66: We have wrappers for casadm, dd, iostat - why not use them?
casadm.set_cache_mode(), test_tools/dd.py, test_tools/iostat.py

test/functional/tests/ci/test_ci_write_mode_read_only.py
```diff
@@ -0,0 +1,187 @@
...
+            write_cache_delta_1 = (int(write_cache_1) - int(write_cache_0)) * iostat_block_size
+            TestRun.executor.run_expect_success(
+                f"dd if=/dev/cas1-1 of=$TMP_DIR/file001.bin bs=4096 count={count} "
+                f"skip=20000 iflag=direct")
+            read_core_2 = TestRun.executor.run_expect_success(
+                f"iostat {core_device.parent_device.path} | grep {tmp_core} | "
+                f"awk 'NR==1 {temp}'").stdout
+            read_cache_2 = TestRun.executor.run_expect_success(
+                f"iostat {cache_device.parent_device.path} | grep {tmp_cache} | "
+                f"awk 'NR==1 {temp}'").stdout
```
Deixx, line 122: wrappers

test/functional/tests/ci/test_ci_write_mode_read_only.py
```diff
@@ -0,0 +1,187 @@
...
+        core_device.create_partitions([Size(2048, Unit.MebiByte)])
+
+        cache_device = cache_device.partitions[0]
+        core_device = core_device.partitions[0]
+
+    with TestRun.step("Test read from cache"):
+        count = 100
+        block_size = 4096
+        iostat_block_size = 1024
+        modes = ["wt", "wa"]
```
Deixx, line 42: I'd prefer using existing enums to strings, which might be longer, but perhaps more maintainable, and if you use casadm wrapper for --set-cache-mode, you can use the enums as args

Deixx (changes requested):

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,54 @@
...
+            f"cmd {third}' | grep -o '{backslash}-{backslash}-{backslash}S*'")
+        commands = output.stdout.splitlines()
+
+        for command in commands:
+            output = TestRun.executor.run_expect_success(
+                f"{casadm_bin} {command} -H | aspell list --lang=en_US "
+                f"--add-extra-dicts={cas_dictionary}")
+
+            if output.stdout:
+                TestRun.LOGGER.error(f"Misspelled word found in command :{command}\n")
```
Deixx, line 53: suggestion ;)
```suggestion
                TestRun.LOGGER.error(f"Misspelled words found in command: {command}\n")
```

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+from test_tools.iostat import IOstatBasic
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.CI()
+def test_ci_write_mode():
+    """
+    title: Verification test for write mode: write around
+    description: Verify if write mode: write around, works as expected and cache only reads
```
Deixx, line 23:
```suggestion
    description: Verify if write around mode works as expected and caches only reads
```

Deixx (dismissed):

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,54 @@
...
+            f"--add-extra-dicts={cas_dictionary}")
+
+        if output.stdout:
+            TestRun.LOGGER.error("Misspelled words found:\n")
+            TestRun.LOGGER.error(output.stdout)
+
+        first = "{ cmd=1;next }"
+        second = "{ cmd=0 }"
+        third = "{ print $0 }"
+        backslash = "\\"
```
robertbaldyga, line 40: Why do we parametrize the command with those fixed strings?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.CI()
+def test_ci_write_mode():
+    """
+    title: Verification test for write mode: write around
+    description: Verify if write around mode works as expected and caches only reads
+    and does not cache write
+    pass criteria:
+    - writes are not cached
```
robertbaldyga, line 26: I guess adding second criterium "reads are cached" may also sound reasonable.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+
+    with TestRun.step("Prepare devices"):
+        cache_device = TestRun.disks["cache"]
+        core_device = TestRun.disks["core"]
+
+        cache_device.create_partitions([Size(1024, Unit.MebiByte)])
+        core_device.create_partitions([Size(2048, Unit.MebiByte)])
```
robertbaldyga, line 34: You could use `Size(1, Unit.GibiByte)` and `Size(2, Unit.GibiByte)` for better readability.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+
+        cache_device = cache_device.partitions[0]
+        core_device = core_device.partitions[0]
+
+    with TestRun.step("Test read from cache"):
+        count = 100
+        block_size = 4096
+        selected_cache_modes = [casadm.CacheMode.WT, casadm.CacheMode.WA]
```
robertbaldyga, line 42: If the test if only for WA, why do we have WT here?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+
+        for mode in selected_cache_modes:
+            print("------Start CAS Linux in " + mode.value + " mode")
```
robertbaldyga, line 45: Please use logger instead. Also these minuses at the beginning are probably not needed.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            # This is where the real test starts
+            cache = casadm.start_cache(cache_dev=cache_device, cache_id=1, force=True)
+            casadm.add_core(cache, core_device)
+            time.sleep(2.5)
+            casadm.set_cache_mode(mode, 1)
```
robertbaldyga, line 50: Why don't you just pass cache mode as a `start_cache()` parameter?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+
+            dd = (
+                Dd()
+                .input("/dev/cas1-1")
+                .output("$TMP_DIR/file001.bin")
```
robertbaldyga, line 55: This temporary file doesn't seem to be needed. Why not use `/dev/null` instead?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+                .count(count)
+                .block_size(Size(1, Unit.Blocks4096))
+                .iflag("direct")
+                .skip(10000)
```
robertbaldyga, line 59: What is that skip for?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            read_cache_delta = (int(read_cache_2) - int(read_cache_1))
+            data_read = count * block_size
+            if read_cache_delta == data_read:
+                TestRun.LOGGER.info(f"Blocks read from cache:{read_cache_delta} == {data_read}")
+            else:
+                TestRun.LOGGER.error(f"Blocks read from cache:{read_cache_delta} != {data_read}")
+
+            casadm.stop_cache(1)
+
+    with TestRun.step("Test write and then read from cache"):
```
robertbaldyga, line 89: This could be make a second test function within this file (e.g. `test_ci_write_around_write`).

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+from storage_devices.disk import DiskType, DiskTypeSet, DiskTypeLowerThan
+from test_utils.size import Unit, Size
+from test_tools.dd import Dd
+from test_tools.iostat import IOstatBasic
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.CI()
+def test_ci_write_mode():
```
robertbaldyga, line 20: The name is very generic. How about `test_ci_write_around_read`?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            if read_cache_delta == data_read:
+                TestRun.LOGGER.info(f"Blocks read from cache:{read_cache_delta} == {data_read}")
+            else:
+                TestRun.LOGGER.error(f"Blocks read from cache:{read_cache_delta} != {data_read}")
+
+            casadm.stop_cache(1)
+
+    with TestRun.step("Test write and then read from cache"):
+        for mode in selected_cache_modes:
+            print("------Start CAS Linux in " + mode.value + " mode")
```
robertbaldyga, line 91: From here onwards pretty the same comments as to the previous one.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+        cache_device = TestRun.disks["cache"]
+        core_device = TestRun.disks["core"]
+
+        cache_device.create_partitions([Size(1024, Unit.MebiByte)])
+        core_device.create_partitions([Size(2048, Unit.MebiByte)])
+
+        cache_device = cache_device.partitions[0]
+        core_device = core_device.partitions[0]
+
+    with TestRun.step("Test read from cache"):
```
robertbaldyga, line 39: This is one huge test step. Please try to split it into few smaller steps so that it will be clear in the log what is going on in the test.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            read_cache_2 = iostat[0].total_reads.value
+
+            read_cache_delta = (int(read_cache_2) - int(read_cache_1))
+            data_read = count * block_size
+            if read_cache_delta == data_read:
+                TestRun.LOGGER.info(f"Blocks read from cache:{read_cache_delta} == {data_read}")
+            else:
+                TestRun.LOGGER.error(f"Blocks read from cache:{read_cache_delta} != {data_read}")
+
+            casadm.stop_cache(1)
```
robertbaldyga, line 87: Please use `cache.stop()` instead.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+                .iflag("direct")
+                .skip(10000)
+            )
+            dd.run()
+
+            iostat = IOstatBasic.get_iostat_list([cache_device.parent_device])
+            read_cache_2 = iostat[0].total_reads.value
+
+            read_cache_delta = (int(read_cache_2) - int(read_cache_1))
+            data_read = count * block_size
```
robertbaldyga, line 81: Why not use `Size(count, Unit.Blocks4096)`? No conversions in the line above would be needed.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            # iostat read and write
+            iostat = IOstatBasic.get_iostat_list([core_device.parent_device])
+            write_core_0 = iostat[0].total_writes.value
+
+            iostat = IOstatBasic.get_iostat_list([cache_device.parent_device])
+            write_cache_0 = iostat[0].total_writes.value
+
+            dd = (
+                Dd()
+                .input("$TMP_DIR/file001.bin")
```
robertbaldyga, line 107: Why not just use `/dev/zero` or `/dev/random`?

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,212 @@
...
+            write_core_1 = iostat[0].total_writes.value
+
+            iostat = IOstatBasic.get_iostat_list([cache_device.parent_device])
+            write_cache_1 = iostat[0].total_writes.value
+
+            iostat = IOstatBasic.get_iostat_list([core_device.parent_device])
+            read_core_1 = iostat[0].total_reads.value
+
+            iostat = IOstatBasic.get_iostat_list([cache_device.parent_device])
+            read_cache_1 = iostat[0].total_reads.value
```
robertbaldyga, line 126: The iostat list is unnecessarily collected twice for each device.

robertbaldyga (changes requested):

robertbaldyga (changes requested):
Please squash these "update*" commits into single commit.

test/functional/tests/cli/test_cli_help_spelling.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_cli_help_spelling():
+    """
+    title: Spelling test for 'help' command
+    description: Validates spelling of 'help' in CLI
+    pass criteria:
+    - no spelling mistakes are found
+    """
+
+    cas_dictionary = os.path.join(TestRun.usr.repo_dir, "test", "smoke_test")
```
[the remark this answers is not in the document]
robertbaldyga: Please move it to `test/functional/resources`.

test/functional/tests/ci/test_ci_write_mode_write_around.py
robertbaldyga: Please use `Udev.enable()` and `Udev.disable()`.

test/functional/tests/ci/test_ci_write_mode_write_around.py
```diff
@@ -0,0 +1,371 @@
...
+        else:
+            TestRun.LOGGER.error(f"Blocks read from cache:{read_cache_delta} != {data_read}")
+
+        cache.stop()
+
+
+@pytest.mark.require_disk("cache", DiskTypeSet([DiskType.nand, DiskType.optane]))
+@pytest.mark.require_disk("core", DiskTypeLowerThan("cache"))
+@pytest.mark.CI()
+def test_ci_write_through_read():
```
mmichal10, line 93: To be honest I think you don't need to add two separate tests for WT and WA. It can be a single test with a properly adapted verification code. You can use trait cache modes API to recognize which cache modes are supposed to insert data on reads and which are not

AdamSeyda: They are two different cases which is why they are separate tests. They also used to be one before the rewrite, which made it overcomplicated and difficult to grasp - separating them was a decision made to improve readability among other things.

mmichal10: It's one test which is supposed to be parametrized by cache modes, actually. Apart from different cache modes, `test_ci_write_through_read` and `test_ci_write_around_read` are exactly the same. I would say leaving them as two separate tests is counterintuitive as a reviewer will be expecting two different scenarios in that case but there is only one. Refactoring them into one test will save time to the reviewers and people investigating the test it if the test fails in the future.

mmichal10 (approved):

kmajzero (approved):

robertbaldyga (approved):

PR #1320, opened 2022-08-25
Tests marking clean up (os-dependend and skipped)

karolinavelkaja:
Marking all os-dependent tests with proper mark: @pytest.mark.os_dependent.

Mark: @pytest.mark.os_dependent_standby_mode_specific - means that those tests should be run only for scopes/clients which are using standby mode functionality (because of their complexity).

Adding mark: @pytest.mark.skip(reason="kedr not supported") - for kedr tests.

### DIFF
--- a/test/functional/tests/cache_ops/test_multistream_seq_cutoff.py
+++ b/test/functional/tests/cache_ops/test_multistream_seq_cutoff.py
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2020-2021 Intel Corporation
+# Copyright(c) 2020-2022 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 #
 import os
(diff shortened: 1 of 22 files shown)


jfckm (approved):

test/functional/tests/incremental_load/test_udev.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2020-2021 Intel Corporation
+# Copyright(c) 2020-2022 Intel Corporation
```
robertbaldyga, line 2: There should not be change in the date when there is no change in the code.

robertbaldyga (changes requested):

karolinavelkaja:
Conflict resolved. One comment resolved. Ready :D

jfckm (approved):

arutk (approved):

PR #1338, opened 2022-09-01
test/api: Add method to create CAS packages

rafalste:

### DIFF
--- a/test/functional/api/cas/installer.py
+++ b/test/functional/api/cas/installer.py
@@ -129,3 +129,33 @@ def check_if_installed(version: str = ""):
         TestRun.LOGGER.info(f"CAS version '{version}' is installed")
 
     return True
+
+
+def create_packages(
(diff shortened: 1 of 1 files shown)


arutk (approved):

test/functional/api/cas/installer.py
```diff
@@ -129,3 +129,33 @@ def check_if_installed(version: str = ""):
...
+        opts += " --debug"
+    if arch:
+        opts += f" --arch {arch}"
+    if source:
+        if package_type.lower() == "rpm":
+            source_package_type = "srpm"
+        elif package_type.lower() == "deb":
+            source_package_type = "dsc"
+        else:
+            source_package_type = ""
```
robertbaldyga, line 156: Shouldn't it be an exception? Maybe in general we can have enum instead of string for package type?

PR #1350, opened 2022-09-14
Add test_io_class_stats_core_cache

anikielx:
Signed-off-by: Adriana Nikelska <adrianax.nikielska@intel.com>


mmichal10:
retest this please

mmichal10:
I would say that testing more than one cache in a single iteration overcomplicates the code whereas the benefits of this approach are negligible. It could be a good idea to simplify it by testing only WT cache mode and leaving only one cache. In that case almost all of my comments become irrelevant

Deixx:
needs modifications: the most important are parametrize by CacheMode and check stats logic

PR #1367, opened 2022-09-29
Workaround for 4.18.0-305.45.1 OpenShift kernel

robertbaldyga:

### DIFF
--- a/configure.d/1_blk_queue_exit_workaround_for_rhel.conf
+++ b/configure.d/1_blk_queue_exit_workaround_for_rhel.conf
@@ -6,14 +6,51 @@
 
 . $(dirname $3)/conf_framework
 
-# RHEL8.5 kernel 4.18 subversion targeted for the workaround is 348.7.1.
-# The workaround is needed starting this version and up.
-QUEUE_PUT_WA_LINUX_MAJOR_SINCE=4
-QUEUE_PUT_WA_LINUX_MINOR_SINCE=18
-QUEUE_PUT_WA_LINUX_SUB_SINCE=0
-QUEUE_PUT_WA_RHEL_MAJOR_SINCE=348
-QUEUE_PUT_WA_RHEL_MINOR_SINCE=7
-QUEUE_PUT_WA_RHEL_SUB_SINCE=1
+# Check whether version A is the same as (or newer than) version B.
+#
+# Version consists of three elements: major, minor and sub.
+# The last argument (GT_LVL) tells at which level it is acceptable
+# for version A to be newer than B. It can be "major", "minor", "sub"
+# or unspecified.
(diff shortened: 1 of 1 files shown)


kmajzero (dismissed):

kmajzero (dismissed):

configure.d/1_blk_queue_exit_workaround_for_rhel.conf
```diff
@@ -15,6 +15,38 @@ QUEUE_PUT_WA_RHEL_MAJOR_SINCE=348
 QUEUE_PUT_WA_RHEL_MINOR_SINCE=7
 QUEUE_PUT_WA_RHEL_SUB_SINCE=1
 
+compare_version () {
```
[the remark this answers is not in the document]
robertbaldyga: Done.

configure.d/1_blk_queue_exit_workaround_for_rhel.conf
```diff
@@ -6,14 +6,51 @@
...
+			  [ "$MINOR_A" -gt "$MINOR_B" ]; then
+			return 0
+		fi
+	elif  [ "$GT_LVL" = "major" ] && [ "$MAJOR_A" -gt "$MAJOR_B" ]; then
+		return 0
+	fi
+	return 1
+}
+
+compare_kernel_version () {
```
arutk, line 48: maybe add the 4-th parameter (GT_LEVEL) so that the interface is the same as compare_rhel_kernel_version?

configure.d/1_blk_queue_exit_workaround_for_rhel.conf
```diff
@@ -24,42 +61,29 @@ check () {
...
 		RHEL_SUB=${arr[2]}
 
-		#check kernel and rhel version (major/minor/sub)
-		if [ "$LINUX_MAJOR" -eq "$QUEUE_PUT_WA_LINUX_MAJOR_SINCE" ] &&
-		   [ "$LINUX_MINOR" -eq "$QUEUE_PUT_WA_LINUX_MINOR_SINCE" ] &&
-		   [ "$LINUX_SUB"   -eq "$QUEUE_PUT_WA_LINUX_SUB_SINCE" ];
-		then
-			if   [ "$RHEL_MAJOR"  -gt "$QUEUE_PUT_WA_RHEL_MAJOR_SINCE" ]; then
+		#check linux kernel version
+		if compare_kernel_version 4 18 0; then
```
arutk, line 73: don't you want to check "==4.18.0" instead of ">= 4.18.0"?

arutk:
As this workaround would break CAS if applied when running on a kernel without the defect, I am still against applying this workaround to unknown future kernel versions. Instead I would prefer to limit the workaround to versions were we know for sure the workaround is required.

PR #1368, opened 2022-09-29
test/api: Add packaging functionality

rafalste:
Needs https://github.com/intel-innersource/frameworks.validation.opencas.test-framework/pull/65

### DIFF
--- a/test/functional/api/cas/packaging.py
+++ b/test/functional/api/cas/packaging.py
@@ -1,0 +1,133 @@
+#
+# Copyright(c) 2022 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+#
+
+
+import os
+import re
+
+from core.test_run import TestRun
+from test_tools.fs_utils import check_if_directory_exists, find_all_files
+from test_tools.packaging import DebSet, RpmSet
+from test_utils.output import CmdException
+
+
+class Packages:
+    # This __init__() method will never be triggered since this class
+    # returns other class in __new__() before __init__() is evaluated.
+    # It is implemented here only to indicate what arguments can be
+    # passed into this class and how are they used.
+    def __init__(self, packages_dir: str = ""):
+        self.packages_dir = packages_dir
+
+    def __new__(cls, *args, **kwargs):
+        distro_id_like = TestRun.executor.run_expect_success(
+            "grep -i ID_LIKE /etc/os-release"
+        ).stdout
+
+        if re.search("rhel|fedora|suse|sles", distro_id_like):
+            return _Rpm(*args, **kwargs)
+        elif re.search("debian", distro_id_like):
+            return _Deb(*args, **kwargs)
+        else:
+            distro_id_like = distro_id_like.split("=")[1].strip("\"'")
+            raise TypeError(f"{distro_id_like} - not recognized OS distribution")
+
+
+class _Rpm(RpmSet):
+    def __init__(self, packages_dir: str = ""):
+        self.packages_dir = packages_dir
+        self.packages = get_packages_list("rpm", self.packages_dir)
+
+    def create(
+        self,
+        sources_dir,
+        packages_dir: str = "",
+        debug: bool = False,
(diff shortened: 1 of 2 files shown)


test/functional/api/cas/packaging.py
```diff
@@ -0,0 +1,136 @@
...
+    packages = []
+    try:
+        for file in find_all_files(packages_dir, recursive=False):
+            # include only binary packages (ready to be processed by package manager)
+            if file.endswith(package_type.lower()) and not file.endswith(
+                "src." + package_type.lower()
+            ):
+                packages.append(os.path.join(packages_dir, file))
+    except (CmdException, ValueError):
+        pass
```
mmichal10, line 106: When the exception may occur? The list of flies is being returned in spite of the loop being interrupted, so what if the list is incomplete? I mean a situation when an exception occurred somewhere in the middle - shouldn't the loop finish its work anyway?

rafalste: find_all_files() raises ValueError when packages_dir is not given - it could be an empty string if not given at the time of constructing Packages() class. And the CmdException can occur when the path given in packages_dir doesn't exist in the system yet - this may happen if we want to create our packages with this class, rather then give the constructor path to already existing packages. In both those cases the returning 'packages' is an empty list and this is expected if we don't want to use existing packages and use create() instead.
Unless there are some unusual circumstances like broken device where 'find' command is stuck during the file search, this CmdException would be raised before 'find' even starts its search and therefore nothing is returned and the 'packages' is still an empty list as expected.
Can you find a scenario where your doubts could be addressed?

mmichal10: How about explicitly checking for value of `packages_dir` and verifying that the given path actually exists? Before the loop even starts. At this moment they sound line hidden assumptions which makes the code harder to understand.

rafalste: The thing is that it is expected that this path may not be given or may not exist. It is expected when you create an instance of the `Packages()` class with the intention to create packages (and then later install and uninstall them), rather than use already existing ones. In this case you won't give the constructor `packages_dir` (if you want to create packages in the default location) or you will (if you want to create them somewhere else) but that directory may not exist yet and will be created during package build.
I previously thought about a different way to handle this by an early return from this method if the `packages_dir` is empty or not exist (before the call to `find_all_files()`), but the way with catching exceptions seems to be more pythonic.
So, what do you think about adding a comment here for a proper explanation? Or let me know if you still stand behind the way with an early return, so I'll make it this way. :)

mmichal10: Btw talking about pythonicness how about using `filter()` function instead of loop?
```
binary_package_filter = lambda file: file.endswith(package_type.lower()) and not file.endswith("src." + package_type.lower())
packages = filter(binary_package_filter , find_all_files(packages_dir, recursive=False))
```

rafalste: > Btw talking about pythonicness how about using filter() function instead of loop?

You're right that this loop is not very pythonic, but even simpler solution then filter() would be a list comprehension. The only drawback of list comprehension in favor of filter() is that filter() returns a generator which uses lazy evaluation, but in this case we still need to convert the resulting filter object to a list, so it makes no difference in the end. Check out the solution that I've pushed just now. :) I think it's even easier to read than the filter() method.

rafalste: > What about this `unusual circumstances like broken device where 'find' command is stuck during the file search`? As this is an invalid scenario, shouldn't an error be somehow propagated to tests? Currently the error is suppressed and the test may fail on a later step

There are many scenarios that can go wrong if something is broken and we are not at all prepared for it in tests. :) But for the sake of simplicity and readability I've added a check for the `packages_dir` instead of catching exceptions.

mmichal10 (approved):

mmichal10 (approved):

karolinavelkaja (approved):

PR #1377, opened 2022-10-10
Open CAS Linux kernel 5.18 support

kmajzero:
This patch introduces necessary changes for OpenCAS Linux to run on Linux kernel v5.18


configure.d/1_bio_alloc.conf
```diff
@@ -0,0 +1,41 @@
...
+    case "$1" in
+    "1")
+		add_function "static inline struct bio *cas_bio_alloc(gfp_t gfp_mask, unsigned short num_vecs)
+		{
+			return bio_alloc(gfp_mask, num_vecs);
+		}" ;;
+    "2")
+		add_function "static inline struct bio *cas_bio_alloc(gfp_t gfp_mask, unsigned short num_vecs)
+		{
+			return bio_alloc(NULL, num_vecs, 0, gfp_mask);
```
robertbaldyga, line 33: Somewhere in `submit_bio_noacct()` (which is called by `submit_bio()`) we have:
```c
struct block_device *bdev = bio->bi_bdev;
struct request_queue *q = bdev_get_queue(bdev);
```
So I suppose we want to set `bdev` to something other than `NULL`.

robertbaldyga (changes requested):

modules/cas_cache/exp_obj.c
```diff
@@ -356,6 +358,10 @@ static void _cas_init_queues(struct cas_disk *dsk)
 	}
 }
 
+#ifndef BLKDEV_DEFAULT_RQ
+#define BLKDEV_DEFAULT_RQ BLKDEV_MAX_RQ
+#endif
```
robertbaldyga, line 363: Looks like something that could be moved to configure.

robertbaldyga (changes requested):

configure.d/1_bd_part_count.conf
```diff
@@ -9,7 +9,11 @@
 check() {
 	cur_name=$(basename $2)
 	config_file_path=$1
-	if compile_module $cur_name "struct gendisk *disk = NULL; struct xarray xa; xa = disk->part_tbl;" "linux/genhd.h"
+	if compile_module $cur_name "struct gendisk *disk = NULL; struct xarray xa; xa = disk->part_tbl;" "linux/blkdev.h"
+	then
+		echo $cur_name "1" >> $config_file_path
+
```
arutk, line 15: looks like it can be combined into a single if with logical "or" in condition, just like in 1_bd_first_part.conf

configure.d/1_bio_clone.conf
```diff
@@ -40,6 +43,12 @@ apply() {
             {
                 return bio_clone_fast(bio, gfp_mask, NULL);
             }" ;;
+    "4")
+        add_function "static inline struct bio *cas_bio_clone(struct bio *bio, gfp_t gfp_mask)
```
arutk, line 47: is "cas_bio_clone" used anywhere? I don't see any reference in my sources.

configure.d/1_get_disk_max_part.conf
```diff
@@ -0,0 +1,33 @@
...
+		echo $cur_name "1" >> $config_file_path
+    	else
+        	echo $cur_name "2" >> $config_file_path
+   	fi
+}
+
+apply() {
+    case "$1" in
+    "1")
+	    	add_define "GET_DISK_MAX_PARTS(x) disk_max_parts(x)"
```
arutk, line 23: "CAS_" prefix would be consistent with other names

configure.d/1_get_disk_max_part.conf
```diff
@@ -0,0 +1,33 @@
...
+
+. $(dirname $3)/conf_framework.sh
+
+check() {
+	cur_name=$(basename $2)
+	config_file_path=$1
+	if compile_module $cur_name "disk_max_parts(NULL)" "linux/bio.h"
+	then
+		echo $cur_name "1" >> $config_file_path
+    	else
```
arutk, line 15: maybe explicitly check whether DISK_MAX_PARTS would compile?

configure.d/1_submit_bio.conf
```diff
@@ -14,18 +14,21 @@
 check() {
 	cur_name=$(basename $2)
 	config_file_path=$1
-	if compile_module $cur_name "submit_bio(NULL);" "linux/bio.h"
+	if compile_module $cur_name "void *t=submit_bio(NULL);" "linux/bio.h"
```
arutk, line 17: submit bio typically returns int status. wouldn't this cause compilation error here when casted to void *?

arutk: Also the return value is never used. Let's just make cas_submit_bio() always have no return value and not differentiate between the two variants of submit_bio (the one returning void and the one returning blk_qc_t)

modules/cas_cache/exp_obj.c
```diff
@@ -161,11 +163,13 @@ static int _cas_del_partitions(struct cas_disk *dsk)
...
-#else
-static int _cas_flags = GENHD_FL_EXT_DEVT;
-#endif
+//#if defined GENHD_FL_NO_PART_SCAN
+//static int _cas_flags = GENHD_FL_NO_PART_SCAN | GENHD_FL_EXT_DEVT;
+//#elif defined GENHD_FL_NO_PART
+static int _cas_flags = GENHD_FL_NO_PART;
+//#else
+//static int _cas_flags = GENHD_FL_EXT_DEV;
+//#endif
```
arutk: doesn't this need to be parametrized in configure?

arutk (dismissed):

modules/cas_cache/exp_obj.c
kmajzero: It already is, yes :)

configure.d/1_bio_clone.conf
kmajzero: removed altogether

robertbaldyga:

configure.d/1_queue_limits.conf
```diff
@@ -15,7 +15,12 @@ check() {
 		echo $cur_name "1" >> $config_file_path
 	elif compile_module $cur_name "struct queue_limits q; q.max_write_zeroes_sectors;" "linux/blkdev.h"
 	then
-		echo $cur_name "2" >> $config_file_path
+		if compile_module $cur_name "struct queue_limits q; q.max_write_same_sectors;" "linux/blkdev.h"
+		then
+			echo $cur_name "2" >> $config_file_path
+		else
+			echo $cur_name "4" >> $config_file_path
+		fi
```
robertbaldyga, line 23: How about changing order so that option 3 goes before 4?

configure.d/1_submit_bio.conf
```diff
@@ -35,26 +25,14 @@ apply() {
...
     "1")
 		add_function "
-	static inline blk_qc_t cas_submit_bio(int rw, struct bio *bio)
+	static inline void cas_submit_bio(int rw, struct bio *bio)
 	{
 		CAS_BIO_OP_FLAGS(bio) |= rw;
-		return submit_bio(bio);
+		submit_bio(bio);
 	}" ;;
+
```
robertbaldyga, line 33: How about kernel versions where `submit_bio()` returns an error code?

configure.d/1_cas_flags.conf
```diff
@@ -0,0 +1,40 @@
...
+        	echo $cur_name 3 >> $config_file_path
+	else
+		echo $cur_name X >> $config_file_path
+    fi
+}
+
+apply() {
+    case "$1" in
+    "1")
+	add_define "_CAS_FLAGS (GENHD_FL_NO_PART_SCAN | GENHD_FL_EXT_DEVT)" ;;
```
robertbaldyga, line 30: This name looks a bit too generic.

modules/cas_cache/volume/vol_block_dev_bottom.c
```diff
@@ -112,27 +113,33 @@ static const char *__block_dev_get_elevator_name(struct request_queue *q)
...
 {
-	if (!q)
+	/*if (!q)
 		return NULL;
 
 	if (q->elevator == NULL)
 		return NULL;
 
 	return __block_dev_get_elevator_name(q);
+	*/
```
robertbaldyga, line 130: Do we need it at all if it always returns NULL?

robertbaldyga (changes requested):

configure.d/1_submit_bio.conf
kmajzero: Most drivers never checked the return value from submit_bio even when submit_bio was declared non-void. It is safe to leave it as it is.

For example, in 4.18, submit_bio is declared as "`blk_qc_t submit_bio(struct bio *bio)`", yet none of the drivers in the ./drivers directory make any use of its return value.

modules/cas_cache/volume/vol_block_dev_bottom.c
kmajzero: Ah, yes. It is going to /dev/null, this one ;)

configure.d/1_cas_flags.conf
kmajzero: changed to _CAS_GENHD_FLAGS

robertbaldyga (changes requested):

robertbaldyga (approved):

robertbaldyga (approved):

KlaudiaJ (approved):

PR #1380, opened 2022-10-11
Use available memory statistic when possible

robertbaldyga:
Update function returning size of free memory to use available statistic if that's supported by given kernel. This function is used to check whether there is enough memory to start cache, and available statistic is preferred for this check over free statistic, as it does include size of page cache allocations which can be easily reclaimed by the system.


rafalste (approved):

configure.d/1_get_free_memory.conf
```diff
@@ -24,15 +27,21 @@ apply() {
     case "$1" in
     "1")
         add_function "
-        static inline unsigned long cas_global_zone_page_state(enum zone_stat_item item)
+        static inline unsigned long cas_get_free_memory(void)
```
arutk, line 30: I suggest changing return type to uint64_t just to be sure

KlaudiaJ (approved):

PR #1402, opened 2022-11-07
tests: Fix invalid mark

robertbaldyga:


mmichal10 (approved):

karolinavelkaja:
As I remember we wanted this 4 tests to be specific - intentionally the mark was different than just "os_dependent". 
This way we will have this "complex" scenarios in regular population of OS dependent list and it was your objection and why new mark was added to distinguish them. What changed? For me it would be fine to change it as long we have common understanding what are the intentions for those scenarios.

katlapinka (approved):

PR #1426, opened 2022-12-28
[v22.6.2] Change CAS_BIO_MAX_VECS value

mmichal10:
Previously when user tried to allocate to big bio he recieved NULL. As the behavior was changed to `BUG()` passing out of range value is no longer legal.

Limiting the requested bio size to BIO_MAX_PAGES (on the older kernels) or to BIO_MAX_VECS (on the newer ones) eliminates the problem as the macros represent the max possible value.

### DIFF
--- a/configure.d/1_bio_max_vecs.conf
+++ b/configure.d/1_bio_max_vecs.conf
@@ -20,9 +20,9 @@ check() {
 apply() {
     case "$1" in
     "1")
-	;;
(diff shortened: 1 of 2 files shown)


robertbaldyga (approved):

mmichal10:
Smoke tests passed on my machine

jfckm (approved):

PR #1470, opened 2024-05-27
Fix compile on Rockylinux 9

MJAsadi72:
Due to custom patches on the RHEL kernel, MAKE_RQ_RET_TYPE uses kernel 5.18.0 and above struct type. this fix prevents using blk_qc_t on Rockylinux 9.

### DIFF
--- a/modules/cas_cache/exp_obj.c
+++ b/modules/cas_cache/exp_obj.c
@@ -31,18 +31,25 @@ static inline void bd_release_from_disk(struct block_device *bdev,
 	return bd_unlink_disk_holder(bdev, disk);
 }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 18, 0)
-	#define KRETURN(x) 	return
-	#define MAKE_RQ_RET_TYPE void
-#elif LINUX_VERSION_CODE >= KERNEL_VERSION(4, 3, 0)
-	#define KRETURN(x)	({ return (x); })
-	#define MAKE_RQ_RET_TYPE blk_qc_t
-#elif LINUX_VERSION_CODE >= KERNEL_VERSION(3, 2, 0)
-	#define KRETURN(x)	return
-	#define MAKE_RQ_RET_TYPE void
+#ifdef RHEL_MAJOR
+       #if RHEL_MAJOR >= 9
+               #define KRETURN(x)      return
+               #define MAKE_RQ_RET_TYPE void
+       #endif
 #else
-	#define KRETURN(x)	({ return (x); })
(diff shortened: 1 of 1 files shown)


mmichal10:
Hello @MJAsadi72,

we appreciate your contribution. Unfortunately, the fix can't be merged it its current form because it breaks compilation on non Red Hat kernels. We're discussing a generic solution. I'll keep you posted on the progress.

Best regards,
Michal

modules/cas_cache/exp_obj.c
```diff
@@ -31,7 +31,7 @@ static inline void bd_release_from_disk(struct block_device *bdev,
 	return bd_unlink_disk_holder(bdev, disk);
 }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 18, 0)
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 18, 0) || RHEL_MAJOR >= 9
```
mmichal10, line 34: @MJAsadi72 could you please test whether the following code works on Rockylinux 9? I've tested it on a non-RHEL kernel and it works fine. 

```suggestion
#ifdef RHEL_MAJOR
       #if RHEL_MAJOR >= 9
               #define KRETURN(x)      return
               #define MAKE_RQ_RET_TYPE void
       #endif
 #else
       #if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 18, 0)
               #define KRETURN(x)      return
               #define MAKE_RQ_RET_TYPE void
       #elif LINUX_VERSION_CODE >= KERNEL_VERSION(4, 3, 0)
               #define KRETURN(x)      ({ return (x); })
               #define MAKE_RQ_RET_TYPE blk_qc_t
       #elif LINUX_VERSION_CODE >= KERNEL_VERSION(3, 2, 0)
               #define KRETURN(x)      return
               #define MAKE_RQ_RET_TYPE void
       #else
               #define KRETURN(x)      ({ return (x); })
               #define MAKE_RQ_RET_TYPE int
       #endif
#endif
```

mmichal10 (approved):

mmichal10:
Fixes #1470

robertbaldyga (approved):

aneagoe:
@robertbaldyga Is it reasonable to expect a new release including this fix shortly, or will that take some time? Looking to upgrade our fleet to EL9 and so far this has been a blocker.

robertbaldyga:
> @robertbaldyga Is it reasonable to expect a new release including this fix shortly, or will that take some time? Looking to upgrade our fleet to EL9 and so far this has been a blocker.

We plan Open CAS v24.09 to be released at the end of September this year.

PR #1481, opened 2024-07-23
Close block devices in a synchronous manner

mmichal10:
fput() doesn't wait for all references on the disk to be unclaimed but instead it only schedules a worker that is supposed to cleanup resources once the device is released.

During cache initialization we open device at least twice - to check its properties and then to actually use it as cache. But since we use the async fput() after the probe, the device might still be in use once we try to open it for the second time (the second open returns -EBUSY).

Using synchronous __fput_sync() to close the device fixes the issue


jfckm (approved):

configure.d/1_bdev_release.conf
```diff
@@ -40,6 +43,9 @@ apply() {
 		add_define "cas_bdev_release(handle, mode, holder) \\
 			bdev_release(handle)" ;;
 	"4")
+		add_define "cas_bdev_release(handle, mode, holder) \\
+			__fput_sync(handle)" ;;
```
robertbaldyga, line 47: `fput()` schedules a delayed work to release the files, so it's just enough to call `reschedule()` just after `fput()` to let the delayed work run.

robertbaldyga (changes requested):

configure.d/1_bdev_release.conf
mmichal10: I've just tested `({ fput(handle); schedule(); })"` and it work (kernel 5.14.0-479.el9.x86_64)

robertbaldyga (approved):

jfckm (approved):

PR #1474, opened 2024-07-10
Implement cache detach/attach

rafalste:

PR #1486, opened 2024-07-26
Mtab check optional

Deixx:
Based on https://github.com/Open-CAS/open-cas-linux/pull/1474 (1 commit added)


casadm/cas_lib.c
```diff
@@ -1803,58 +1803,61 @@ int add_core(unsigned int cache_id, unsigned int core_id, const char *core_devic
...
+		if (core_id < 0) {
+			cas_printf(LOG_ERR,
+				   "Can't stop cache instance %d. Device %s is mounted!\n",
+				   cache_id, mstruct->mnt_fsname);
+		} else {
+			cas_printf(LOG_ERR,
+				   "Can't remove core %d from cache %d."
+				   " Device %s is mounted!\n",
+				   core_id, cache_id, mstruct->mnt_fsname);
 			}
```
mmichal10, line 1845: This function should only check whether cores are mounted. The error messages should be printed by the caller function

mmichal10 (changes requested):

PR #1504, opened 2024-09-06
Update needed RAM calculation method

robertbaldyga:


modules/cas_cache/layer_cache_management.c
```diff
@@ -2037,27 +2037,45 @@ static void init_instance_complete(struct _cache_mngt_attach_context *ctx,
...
+
+	volume_size = ocf_volume_get_length(ctx->device_cfg.volume);
+	ctx->min_free_ram = ocf_mngt_get_ram_needed(cache, volume_size);
+	ocf_volume_close(ctx->device_cfg.volume);
+
+destroy_config:
+	cache_mngt_destroy_cache_device_cfg(&ctx->device_cfg);
+end:
+	if (result)
+		printk(KERN_WARNING "Cannot calculate amount of DRAM needed\n");
```
mmichal10, line 2067: Doesn't it sound a bit clumsy? 😅 "**Failed** to calculate the required amount of DRAM"?

mmichal10 (approved):

modules/cas_cache/layer_cache_management.c
robertbaldyga: Technically it did not fail to calculate it - the calculation itself is trivial and it's guaranteed to succeed. What happened here is that some prerequisites to the calculation were not satisfied, so the calculation was not possible at all. Therefore **cannot** to emphasise this inability to calculate.

rafalste (approved):

PR #1514, opened 2024-09-10
Handle missing /etc/mtab and modify output

Deixx:
resolves https://github.com/Open-CAS/open-cas-linux/issues/1515

Output after modifications:
```
# lsblk
...
vdc          disk
└─cas1-1     disk /tmp/mnt2
vdd          disk
└─cas1-2     disk
  ├─cas1-2p1 part
  ├─cas1-2p2 part /tmp/mnt1
  ├─cas1-2p3 part
  └─cas1-2p4 part /tmp/mnt3
# casadm -L
type    id   disk       status    write policy   device
cache   1    /dev/vdb   Running   wt             -
├core   1    /dev/vdc   Active    -              /dev/cas1-1
└core   2    /dev/vdd   Active    -              /dev/cas1-2
# casadm -T -i 1
Can't stop cache instance 1 due to mounted devices:
/dev/cas1-2p2
/dev/cas1-1
/dev/cas1-2p4
# casadm -R -i 1 -j 1
Can't remove core 1 from cache 1 due to mounted devices:
/dev/cas1-1
# casadm -R -i 1 -j 2
Can't remove core 2 from cache 1 due to mounted devices:
/dev/cas1-2p2
/dev/cas1-2p4
```

### DIFF
--- a/casadm/cas_lib.c
+++ b/casadm/cas_lib.c
@@ -45,8 +45,8 @@
 
 #define CORE_ADD_MAX_TIMEOUT 30
 
-int is_cache_mounted(int cache_id);
-int is_core_mounted(int cache_id, int core_id);
+bool cache_mounts_detected(int cache_id);
+bool core_mounts_detected(int cache_id, int core_id);
 
 /* KCAS_IOCTL_CACHE_CHECK_DEVICE  wrapper */
 int _check_cache_device(const char *device_path,
@@ -1119,7 +1119,7 @@ int stop_cache(uint16_t cache_id, int flush)
 	int status;
 
 	/* Don't stop instance with mounted filesystem */
-	if (is_cache_mounted(cache_id) == FAILURE)
+	if (cache_mounts_detected(cache_id))
 		return FAILURE;
 
 	fd = open_ctrl_device();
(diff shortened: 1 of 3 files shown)


mmichal10 (approved):

test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
```diff
@@ -104,6 +106,10 @@ def test_stop_cache_with_mounted_partition():
         core = cache.add_core(core_part)
         core.mount(mount_point)
 
+    with TestRun.step("Ensure /etc/mtab exists."):
+        if not fs_utils.check_if_symlink_exists("/etc/mtab"):
+            Symlink.create_symlink("/proc/self/mounts", "/etc/mtab")
```
rafalste, line 127: Are you sure every distro links mtab to /proc/self/mounts ? :)

casadm/cas_lib.c
```diff
@@ -1825,38 +1824,39 @@ bool _device_mounts_detected(int cache_id, int core_id)
...
+	mtab = setmntent("/etc/mtab", "r");
+	if (!mtab) {
+		/* should exist, but if /etc/mtab not found we cannot print mounted devices */
+		return;
+	}
+
+	while ((mstruct = getmntent(mtab)) != NULL) {
+		error = strcmp_s(mstruct->mnt_fsname, cmplen, pattern, &no_match);
+		/* mstruct->mnt_fsname is /dev/... block device path, not a mountpoint */
+		if (error != EOK || no_match)
```
rafalste, line 1855: Again, the error here gets ignored.

casadm/cas_lib.c
```diff
@@ -1119,8 +1119,16 @@ int stop_cache(uint16_t cache_id, int flush)
 	int status;
 
 	/* Don't stop instance with mounted filesystem */
-	if (cache_mounts_detected(cache_id))
+	int cmplen = 0;
+	char pattern[80];
+
+	/* verify if any core (or core partition) for this cache is mounted */
+	cmplen = snprintf(pattern, sizeof(pattern), "/dev/cas%d-", cache_id) - 1;
```
rafalste, line 1126: Again, removing null from the end, which shouldn't be there...?

casadm/cas_lib.c
```diff
@@ -1866,7 +1866,22 @@ int remove_core(unsigned int cache_id, unsigned int core_id,
...
 
 	/* don't even attempt ioctl if filesystem is mounted */
-	if (core_mounts_detected(cache_id, core_id)) {
+	bool mounts_detected = false;
+	int cmplen = 0;
+	char pattern[80];
+
+	/* verify if specific core is mounted */
+	cmplen = snprintf(pattern, sizeof(pattern), "/dev/cas%d-%d", cache_id, core_id);
+	if (!(mounts_detected = device_mounts_detected(pattern, cmplen))) {
```
rafalste, line 1875: I would split this into a separate lines with assignment and then check, just for readability.

mmichal10 (dismissed):

casadm/cas_lib.c
```diff
@@ -1803,58 +1803,60 @@ int add_core(unsigned int cache_id, unsigned int core_id, const char *core_devic
...
 	if (core_id >= 0) {
 		/* verify if specific core is mounted */
-		snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-%d", cache_id, core_id);
+		cmplen = snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-%d", cache_id, core_id);
 	} else {
-		/* verify if any core from given cache is mounted */
-		snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-", cache_id);
+		/* verify if any core from given cache is mounted
+		   do not compare terminating NULL for cache */
+		cmplen = snprintf(dev_buf, sizeof(dev_buf), "/dev/cas%d-", cache_id) - 1;
```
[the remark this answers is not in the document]
Deixx: This subtraction was a result of tests - without it the match was exact and this was the reason why previously we always got the "kernel version" of the messages (`Device mounts are pending`) when stopping cache - casadm couldn't find the exact match for `/dev/casX-`, so the check always fell back to the kernel.

casadm/cas_lib.c
```diff
@@ -1803,58 +1803,60 @@ int add_core(unsigned int cache_id, unsigned int core_id, const char *core_devic
...
 	}
 
 	mtab = setmntent("/etc/mtab", "r");
-	if (!mtab)
-	{
-		cas_printf(LOG_ERR, "Error while accessing /etc/mtab\n");
-		return FAILURE;
+	if (!mtab) {
+		/* if /etc/mtab not found then the kernel will check for mounts */
+		return false;
```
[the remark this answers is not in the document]
Deixx: If we can identify that the device is mounted, then we don't even try removing/stopping it, otherwise we try executing operation and return external error if it fails. For this particular case - line 1915 prints the error message and the external error (`Device mounts are pending`).

casadm/cas_lib.c
```diff
@@ -1803,58 +1803,60 @@ int add_core(unsigned int cache_id, unsigned int core_id, const char *core_devic
...
-					   "Can't stop cache instance %d. Device %s is mounted!\n",
-					   cache_id, mstruct->mnt_fsname);
-			} else {
-				cas_printf(LOG_ERR,
-					   "Can't remove core %d from cache %d."
-					   " Device %s is mounted!\n",
-					   core_id, cache_id, mstruct->mnt_fsname);
-			}
-			return FAILURE;
+			return false;
```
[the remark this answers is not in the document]
Deixx: That's right - this is the main and needed change here. `false` indicates that we did not or could not detect any mounts for whatever reason (hence the change of the function name), and that we should proceed with the operation and let it fail at a later stage.

test/functional/tests/fault_injection/test_fault_injection_with_mounted_core.py
Deixx: I'm not - I'll change it so this checks if any type of /etc/mtab exists. If it does we'll leave it alone, if not then creating such a symlink shouldn't break anything ;)

casadm/cas_lib.c
Deixx: If we get an error, then we cannot determine if the device is mounted or not -> let's at least try to remove it.

casadm/cas_lib.c
Deixx: I stand by the version that it should :)

rafalste (approved):

PR #1536, opened 2024-09-26
Update test_4k_100p_hit_reads_wt

Deixx:
Two main changes:
- spread jobs over device area equal to core size (before some jobs were configured outside of core and skipped)
- divided workload equally among jobs (before every consecutive job run bigger workload then the last)

Refactor

### DIFF
--- a/test/functional/tests/performance/test_100p_hits.py
+++ b/test/functional/tests/performance/test_100p_hits.py
@@ -1,5 +1,6 @@
 #
 # Copyright(c) 2020-2022 Intel Corporation
+# Copyright(c) 2024 Huawei Technologies
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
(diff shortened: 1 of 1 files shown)


test/functional/tests/performance/test_100p_hits.py
```diff
@@ -31,14 +32,17 @@
...
-        title: Test CAS performance in 100% Cache Hit scenario
-        description: |
-          Characterize cache device with workload (parametrized by qd and job number), and then run
-          the same workload on cached volume.
-        pass_criteria:
-          - always passes
+    title: Test CAS performance in 100% Cache Hit scenario
+    description: |
+      Characterize cache device with workload (parametrized by qd and job number), and then run
+      the same workload on cached volume.
```
Kamoppl, line 38:
```suggestion
        Characterize cache device with workload (parametrized by qd and job number), and then run
        the same workload on cached volume.
```

Kamoppl (approved):

test/functional/tests/performance/test_100p_hits.py
```diff
@@ -49,24 +53,47 @@ def test_4k_100p_hit_reads_wt(queue_depth, numjobs, cache_line_size, perf_collec
         .io_depth(queue_depth)
         .cpus_allowed(get_dut_cpu_physical_cores())
         .direct()
+        .io_size(size_per_job)
     )
+    # spread jobs
+    offset = (testing_range / numjobs).align_down(Unit.Blocks512.value)
```
rafalste, line 59: Is the testing_range correct for calculating the offset? testing_workset value was used here before.

rafalste (commented):
Also disable_wbt_throttling() function is not used.

test/functional/tests/performance/test_100p_hits.py
Deixx: testing_workset is larger than the core, so part of the offsets was actually outside of core bounds, and the jobs were ignored by fio

PR #1543, opened 2024-10-02
Update for test_multistream_seq_cutoff_functional

Deixx:
Throw exception when dd runs out of space
Shorter logs
Blocks instead of bytes in error message

### DIFF
--- a/test/functional/tests/cache_ops/test_multistream_seq_cutoff.py
+++ b/test/functional/tests/cache_ops/test_multistream_seq_cutoff.py
@@ -6,16 +6,18 @@
 
 import posixpath
 import random
-import pytest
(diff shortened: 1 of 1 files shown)


test/functional/tests/cache_ops/test_multistream_seq_cutoff.py
```diff
@@ -6,23 +6,26 @@
 
 import posixpath
 import random
-import pytest
-
 from time import sleep
 
+import pytest
```
mmichal10, line 11: 🚀

mmichal10 (approved):

Kamoppl (approved):

PR #1575, opened 2024-10-23
add random operation test

Kamoppl:


test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,501 @@
...
+            "used_cache_devices": used_cache_devices,
+            "available_cache_devices": available_cache_devices,
+            "used_core_devices": used_core_devices,
+            "available_core_devices": available_core_devices,
+        }
+        operation_list = [
+            start_cache,
+            stop_cache,
+            set_param,
+            set_param,
```
Deixx, line 85:
```suggestion
            set_param,
```

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,501 @@
...
+
+def stop_cache(devices: dict) -> str | None:
+    available_cache_devices = devices.get("available_cache_devices")
+    used_cache_devices = devices.get("used_cache_devices")
+    available_core_devices = devices.get("available_core_devices")
+    used_core_devices = devices.get("used_core_devices")
+    cache_list = get_caches()
+
+    if not cache_list:
+        return
```
Deixx, line 177:
```suggestion
        return None
```

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+def start_cache(devices: dict) -> str | None:
+    available_cache_devices = devices.get("available_cache_devices")
+    used_cache_devices = devices.get("used_cache_devices")
+    available_core_devices = devices.get("available_core_devices")
+    used_core_devices = devices.get("used_core_devices")
+
+    if not available_cache_devices:
+        return None
```
Deixx, line 119: make returns consistent (either just `return` or `return None`)

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,501 @@
...
+        used_cache_devices = [cache_dev.partitions[0]]
+        available_cache_devices = cache_dev.partitions[1:]
+        used_core_devices = [core_dev.partitions[0]]
+        available_core_devices = core_dev.partitions[1:]
+        devices = {
+            "used_cache_devices": used_cache_devices,
+            "available_cache_devices": available_cache_devices,
+            "used_core_devices": used_core_devices,
+            "available_core_devices": available_core_devices,
+        }
```
Deixx, line 80: I really don't really like the idea of putting the lists in a dict and in each operation function getting them back using the string keys. 
Alternatives:
- make the lists global
- define the operation functions within the test so they can access those lists directly
- make a class with the lists as variables  

Preferably one of the first two.

Deixx (commented):
Overall looks good - minor suggestions

Deixx (approved):

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,545 @@
...
+                    f"Threshold: {str(random_threshold)}\n"
+                    f"Policy: {str(random_policy)}\n"
+                    f"Threshold: {str(random_threshold)}\n"
+                )
+            else:
+                core_list = cache_to_set_param.get_core_devices()
+                if not core_list:
+                    return None
+                random_core = random.choice(core_list)
+                random_threshold = Size(random.randint(1, 4194181))
```
katlapinka, line 264: All value ranges for configuration parameters should be placed somewhere as constants, to be able to update it easily in case of any requirement changes. It can be done in a separate PR, I am leaving this comment just as the reminder.

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,501 @@
...
+        step_name_list = []
+
+        try:
+            while datetime.now() < test_start_time + test_runtime:
+                step = random.choice(operation_list)
+                step_name = step(devices)
+                if step_name:
+                    step_name_list.append(step_name)
+        except CmdException:
+            step_name_list.append(step_name)
```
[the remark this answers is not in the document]
Kamoppl: These functions do not have the step.__name__ method, and they return strings describing exactly what happened inside each step.

katlapinka (approved):

test/functional/tests/misc/test_random_cache_ops.py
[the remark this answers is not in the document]
Deixx: Maybe the steps should have some basic description - if there's an exception during a step execution, then what does this append? The result step_name of the previous step?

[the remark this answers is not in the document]
Kamoppl: fine, you won

test/functional/tests/misc/test_random_cache_ops.py
```diff
@@ -0,0 +1,632 @@
...
+                step_name = step(
+                    available_cache_devices,
+                    used_cache_devices,
+                    available_core_devices,
+                    used_core_devices,
+                )
+                if step_name:
+                    step_name_list.append(step.__name__)
+        except CmdException:
+            step_name_list.append(step.__name__)
```
Deixx, line 110: I meant the step.\_\_name\_\_ only for the exception handling - that is the case when step() throws an Exception and doesn't return anything (so step_name is still set by the previous step)
```suggestion
                if step_name:
                    step_name_list.append(step_name)
        except CmdException:
            step_name_list.append(step.__name__)
```

Deixx (approved):

katlapinka (approved):

PR #1619, opened 2025-02-28
New IO class rule `io_direction`

Deixx:
Doc update https://github.com/Open-CAS/open-cas.github.io/pull/92

### DIFF
--- a/modules/cas_cache/classifier.c
+++ b/modules/cas_cache/classifier.c
@@ -221,6 +221,42 @@ static int _cas_cls_string_ctr(struct cas_classifier *cls,
 	return 0;
 }
 
+/* IO direction condition constructor. @data is expected to contain string
+ * translated to IO direction.
+ */
+static int _cas_cls_direction_ctr(struct cas_classifier *cls,
+		struct cas_cls_condition *c, char *data)
+{
+	uint64_t direction;
+	struct cas_cls_numeric *ctx;
+
+	if (!data) {
+		CAS_CLS_MSG(KERN_ERR, "Missing IO direction specifier\n");
+		return -EINVAL;
+	}
+
+	if (strncmp("read", data, 5) == 0) {
+		direction = READ;
+	} else if (strncmp("write", data, 6) == 0) {
+		direction = WRITE;
+	} else {
+		CAS_CLS_MSG(KERN_ERR, "Invalid IO direction specifier '%s'\n"
+			" allowed specifiers: 'read', 'write'\n", data);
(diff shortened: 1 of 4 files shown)


configure.d/1_bio_flags.conf
```diff
@@ -33,6 +34,9 @@ apply() {
     *)
         exit 1
     esac
+	
+	add_define "CAS_BIO_DIRECTION(bio) \\
+			((CAS_BIO_OP_FLAGS(bio)) & REQ_OP_WRITE)"
```
robertbaldyga, line 39: There is `bio_data_dir(bio)`.

modules/cas_cache/classifier.c
```diff
@@ -221,6 +221,42 @@ static int _cas_cls_string_ctr(struct cas_classifier *cls,
...
+
+	ctx = kmalloc(sizeof(*ctx), GFP_KERNEL);
+	if (!ctx)
+		return -ENOMEM;
+	ctx->operator = cas_cls_numeric_eq;
+
+	if (strncmp("read", data, len) == 0) {
+		ctx->v_u64 = REQ_OP_READ;
+	} else if (strncmp("write", data, len) == 0) {
+		ctx->v_u64 = REQ_OP_WRITE;
```
robertbaldyga, line 248: `REQ_OP_READ`/`REQ_OP_WRITE` are technically correct in this context, however in other places in code we use `READ`/`WRITE`, which are more generic Linux kernel macros for data direction. I think we may want to keep it consistent, one way or another.

robertbaldyga (changes requested):

robertbaldyga (dismissed):

modules/cas_cache/classifier.c
```diff
@@ -221,6 +221,42 @@ static int _cas_cls_string_ctr(struct cas_classifier *cls,
...
+
+	if (!data) {
+		CAS_CLS_MSG(KERN_ERR, "Missing IO direction specifier\n");
+		return -EINVAL;
+	}
+
+	len = strnlen(data, MAX_STRING_SPECIFIER_LEN);
+
+	ctx = kmalloc(sizeof(*ctx), GFP_KERNEL);
+	if (!ctx)
```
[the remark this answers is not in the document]
Deixx: No other ENOMEM in this file has a specific message - isn't there some generic message for that?

mmichal10 (approved):

robertbaldyga (approved):

PR #1623, opened 2025-03-14
Involuntary preemption

mmichal10:
Prevent loading the kernel module if the kernel can be involuntarily preempted


Deixx (approved):

modules/cas_cache/main.c
```diff
@@ -42,10 +43,52 @@ MODULE_PARM_DESC(seq_cut_off_mb,
...
+
+	return false;
+}
+
 static int __init cas_init_module(void)
 {
 	int result = 0;
 
+	if (involuntary_preemption_enabled())
+		return -EINVAL;
```
robertbaldyga, line 90: `return -ENOTSUP;`

modules/cas_cache/main.c
```diff
@@ -42,10 +43,52 @@ MODULE_PARM_DESC(seq_cut_off_mb,
...
+				"involuntary preemption\n");
+
+		return true;
+	}
+
+#ifdef CONFIG_PREEMPT_DYNAMIC
+	/* preempt_model_none() or preempt_model_voluntary() are not defined if
+	 * the kernel has been compiled without PREEMPT_DYNAMIC
+	 */
+	printk(KERN_ERR OCF_PREFIX_SHORT
```
robertbaldyga, line 72: Why do we print this message at the `KERN_ERR` verbosity even if the preemption is set to voluntary?

robertbaldyga (changes requested):

mmichal10:
Regarding checkpatch's `WARNING: else is not generally useful after a break or return` - in this particular case I consider the else statement quite useful

Deixx (dismissed):

mmichal10:
Btw that's how the Makefile reports the error:
obraz
I think we might want to change it for sth more descriptive

modules/cas_cache/main.c
```diff
@@ -42,10 +43,59 @@ MODULE_PARM_DESC(seq_cut_off_mb,
...
+	if (!config_dynamic && !config_rt && !config_preempt && !config_lazy)
+		return false;
+
+	if (config_none)
+		return false;
+
+	if (config_rt || config_preempt || config_lazy) {
+		printk(KERN_ERR OCF_PREFIX_SHORT
+				"Open CAS doesn't support kernels with "
+				"involuntary preemption\n");
```
robertbaldyga, line 63: Maybe, to make both error messages to have kind of similar format?
```c
"The kernel has been built with involuntary preemption "
"enabled.\nFailed to load Open CAS kernel module.\n"
```

robertbaldyga (changes requested):

Deixx (approved):

robertbaldyga (approved):

Deixx (approved):

PR #1638, opened 2025-03-28
tests: Remove serial number check

jfckm:


Kamoppl (approved):

Deixx:
What's wrong with the check?

katlapinka (commented):
I think this check helps to prevent using unwanted disk from the DUT. What was the purpose for removing it?

PR #1787, opened 2026-08-03
fix(dkms): run ./configure on the correct kernel headers

### LINKED ISSUE #1772: Support linux 7.0
## Description
Support for building with linux 7.0

## Reason
Proxmox VE 9.2 depends on Linux 7.0 so OpenCAS is not usable with it

kaysond:
Fixes #1772 

This was written by Claude Opus, so it's a little aggressive in preventing the issue from occurring, as coding agents tend to be defensive at paranoid levels. I figured I'd leave it all here in case any of it is desirable for the team.

Please feel free to leave feedback then I'll clean up/update the PR manually.

### DIFF
--- a/configure
+++ b/configure
@@ -110,6 +110,9 @@ fi
 
 if [ ! -f /usr/include/libelf.h ]; then
 	KERNEL_CONFIG="$KERNEL_DIR/.config"
+	if [ ! -f "$KERNEL_CONFIG" ]; then
+		KERNEL_CONFIG="$KERNEL_DIR/include/config/auto.conf"
+	fi
 	if [ -f "$KERNEL_CONFIG" ]; then
 		if grep ^CONFIG_UNWINDER_ORC=[Yy] "$KERNEL_CONFIG" &>/dev/null; then
 			echo "Error: CONFIG_UNWINDER_ORC=y option is set in your kernel,"\
--- a/configure
+++ b/configure
@@ -146,6 +146,11 @@ FILES_COUNT=`echo $CONFIG_FILES | wc -w`
 
 CONFIG_FILE=$SCRIPTPATH/"config.out"
 
+# Name under which the probed kernel is stored in the config file, so that a
+# header generated from a saved config still states which kernel API it
+# describes - see generate_header() and modules/Makefile.
+KERNEL_DIR_TAG="CONFIGURED_KERNEL_DIR"
+
 generate_config() {
 	rm -f ${CONFIG_FILE}
 	touch ${CONFIG_FILE}
@@ -175,6 +180,8 @@ generate_config() {
 		grep "X" ${CONFIG_FILE} | cut -f1 -d ' '
 		exit 1
 	fi
+
+	echo "$KERNEL_DIR_TAG $(realpath "$KERNEL_DIR")" >> ${CONFIG_FILE}
 }
 
 generate_header() {
@@ -185,6 +192,12 @@ generate_header() {
 
 	echo "#ifndef __GENERATED_DEFINES_H__" >> $SCRIPTPATH/modules/generated_defines.h
 	echo "#define __GENERATED_DEFINES_H__" >> $SCRIPTPATH/modules/generated_defines.h
+
+	CONFIGURED_KERNEL_DIR=$(grep "^$KERNEL_DIR_TAG " ${CONFIG_FILE} | cut -d' ' -f2-)
+	if [ -n "$CONFIGURED_KERNEL_DIR" ]; then
+		echo "#define CAS_CONFIGURED_KERNEL_DIR \"$CONFIGURED_KERNEL_DIR\"" \
+			>> $SCRIPTPATH/modules/generated_defines.h
+	fi
 
 	echo "Configuring OpenCAS"
 	for file in $FIRST; do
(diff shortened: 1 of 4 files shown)


configure
```diff
@@ -146,14 +230,28 @@ generate_header() {
 	echo "#endif /* __GENERATED_DEFINES_H__ */" >> $SCRIPTPATH/modules/generated_defines.h
 }
 
+# Records which kernel the generated header was probed against, so that a build
+# aimed at a different one fails with a clear message instead of a wall of
+# compiler errors - see modules/Makefile.
+KERNEL_STAMP=$SCRIPTPATH/modules/.configured_kernel
```
robertbaldyga: The problem with having a separate file for this is apparent a few lines below: when using an existing config file, we cannot perform the check. The fix is very simple - include the kernel path in the `config.out` instead of introducing a separate file.

It's also worth adding it as a define or comment to `generated_defines.h`, so that `make` does not need to know the path to the config (in case it was not `config.out`).

modules/Makefile
```diff
@@ -80,6 +99,6 @@ uninstall:
 
 reinstall: uninstall install
 
-.PHONY: all default clean distclean sync distsync install uninstall
+.PHONY: all default check_configured_kernel clean distclean sync distsync install uninstall
```
robertbaldyga: This is a nice safety mechnism. Not related to the DKMS fix directly though, so I'd suggest moving it to a separate commit.

configure
```diff
@@ -7,6 +7,87 @@
...
+
+# Pick the kernel to configure for: the options win over the environment, and
+# a directory wins over a version. With nothing given at all, fall back to the
+# running kernel - the same default configure.d/conf_framework.sh and
+# modules/Makefile use.
+if [ -n "$OPT_KERNEL_DIR" ]; then
+	KERNEL_DIR="$OPT_KERNEL_DIR"
+elif [ -n "$OPT_KERNEL_VERSION" ]; then
+	KERNEL_DIR="/lib/modules/$OPT_KERNEL_VERSION/build"
+elif [ -z "$KERNEL_DIR" ] && [ -n "$KERNEL_VERSION" ]; then
```
robertbaldyga: Having multiple ways to set `KERNEL_DIR` and `KERNEL_VERSION` is asking for problems. We can simply do `else` here and hard overwrite the variables to avoid confusion.

configure
```diff
@@ -7,6 +7,87 @@
...
+	KERNEL_DIR="/lib/modules/$KERNEL_VERSION/build"
+fi
+KERNEL_DIR="${KERNEL_DIR:-/lib/modules/$(uname -r)/build}"
+export KERNEL_DIR
+
```
robertbaldyga, line 60: This entire `./configure` paremeters rework deserves a separate commit.

configure
```diff
@@ -58,7 +139,10 @@ then
 fi
 
 if [ ! -f /usr/include/libelf.h ]; then
-	KERNEL_CONFIG=$(find /usr/src/ -type d -name "*$(uname -r)")/.config
+	KERNEL_CONFIG="$KERNEL_DIR/.config"
+	if [ ! -f "$KERNEL_CONFIG" ]; then
+		KERNEL_CONFIG="$KERNEL_DIR/include/config/auto.conf"
+	fi
```
robertbaldyga, line 115: This part should be a yet another separate commit.

robertbaldyga (changes requested):

configure
robertbaldyga: When dropping setting the `KERNEL_DIR` by env variable, we also need to update it in RPM building path (which is a single line change).

robertbaldyga:
The overall idea behind those changes is great, but as it goes with LLM generated code, there is a huge imbalance between paranoid defensiveness and corner case blindness :)

kaysond:
@robertbaldyga - I took a look at everything and updated the build setup and commit history as requested.

1. adds a single kernel dir arg to ./configure (the same commit updates the rpm build so we dont break the build now that the env var isnt used at all). 
2. fix the kernel config path. you requested a separate commit for this, so that's what I did. I think it should be squashed into the first, though, because not having it is breaking
3. fix dkms
4. add the configure/make kernel version consistency check. not having this doesn't strictly break anything, so i think it's fine to add it at the end

Let me know if you have any more feedback, esp. on whether I should squash the first two commits together

kaysond:
bump. @robertbaldyga - any feedback?

configure
robertbaldyga: Ah, I see what you mean by saying it should be part of the first commit. So, in this diff there are actualy two distinct changes:
1. Change the .config path from uname-based to $KERNEL_CONFIG-based (line 112). That one indeed should belong to the first commit, otherwise the change is not complete.
2. Add /include/config/auto.conf as a fallback. This is not directly related wtih ./configure parameter rework, so I'd keep it as a separate commit.

robertbaldyga (changes requested):

robertbaldyga:
> bump. @robertbaldyga - any feedback?

One small comment + please look at the checkpatch output.
The ruff seems to return false positives, so ignore it for now - we will try to fix it soon.

robertbaldyga (approved):

PR #1788, opened 2026-08-06
rpm: build kernel modules via DKMS on the target

Arondight:
Switches RPM packaging from prebuilt kmod modules (one per-kernel subpackage, built on the host) to DKMS source built on the target at install time. The build host no longer needs kernel-devel. opencas_exporter is split into an optional subpackage (--with-exporter, off by default) so the default build doesn't pull golang.

Tested on multi kernels: fresh install, clean uninstall, dkms→dkms upgrade, multi-kernel build, kmod→dkms upgrade, kmod→dkms + uninstall, reinstall. All Pass.


robertbaldyga:
@Arondight Hi 秦凡东! Thank you for this contribution. It took a while for me to find a moment to look at it. I let myself to make a few modifications, as many users currently rely on the existing RPM building method and they don't want to switch to DKMS. Therefore I made DKMS RPM a second option, so that the users have a choice of which one they want to build. I also made sure that transitions in both direction (prebuilt->DKMS and DKMS->prebuilt) work well. I'll publish my tests once I get them polished.

